PR #5, opened 2020-04-07
Fixing Integration Tests and Bugs

robertbartel:
An initial PR, not yet complete, with fixes to get existing integration tests working.  

## Changes

- Fixes to the logic and design of several integration tests, accounting for recent deliberate code changes
- Fixes to several bugs in the main package logic that had been introduced, as revealed by the tests

## Notes

- While the output is not clean, the tests relying on a secondary subprocess to have a listener class be tested are functioning (though this may need to be improved)
- For the `scheduler` package in particular, and the RedisManager class, I get different sets of results to tests when I call the tests via the `test_package.sh` script than when I run the tests via my IDE (having already manually run `setup_it_env.sh` to start the Redis container)
    - I am still investigating what the difference is and the full extent of the implications

## Todos

- ~The way serialized objects are deserialized will have to be changed, perhaps largely redesigned, as pipelines were not playing nice (see TODO comments where this got done)~ This has been fixed now with subsequent commits.


python/lib/scheduler/dmod/scheduler/resources/redis_manager.py
```diff
@@ -26,8 +26,9 @@ class RedisManager(ResourceManager, RedisBacked):
 
     def __init__(self, resource_pool: str, redis_host: Optional[str] = None, redis_port: Optional[int] = None,
                  redis_pass: Optional[str] = None, **kwargs):
-        super().__init__(resource_pool=resource_pool, redis_host=redis_host, redis_port=redis_port,
-                         redis_pass=redis_pass, **kwargs)
+        super().__init__(redis_host=redis_host, redis_port=redis_port, redis_pass=redis_pass, **kwargs)
```
hellkite500, line 29: Just ran into this bug myself

hellkite500:
Is this still UNFINISHED? Or is it in a state we can merge?

robertbartel:
This change set can be considered finished, so I'll merge them in now.

PR #39, opened 2020-09-08
Testing improvements and fixes

robertbartel:
Fixes to dependencies and helper scripts to make sure testing is successful.

* improving self-signed cert helper script
* making sure integration test helper scripts set up everything in environment as needed
* bumping library package versions and dependency versions


hellkite500 (approved):

hellkite500:
Note that since we don't maintain older package versions, we basically have to bump every requirements.txt/setup.py when we bump a version.

PR #45, opened 2020-09-29
Generalize launcher

hellkite500:
[Short description explaining the high-level reason for the pull request]

## Additions

- add `domain` property to request

## Removals

-

## Changes

- refactor `image_and_domain.yml` file for better support of multiple models
- expose `version` as an instance property instead of `@classmethod`
- set service name based on job model name
- change NWM model name to lowercase nwm
- initialize all `JobImpl` objects with a random uuid using `uuid.uuid4()`

## Testing

1. Existing tests pass, start_job tested with ngen and nwm

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Other

- [ ] No linting errors or warnings

PR #46, opened 2020-09-29
Ngen model support

hellkite500:
Adds support for a NGEN request and job launching
## Additions

- NGENRequest
- NGENREquestResponse
- NGEN Docker Container entry point

## Changes

- Fixed initialization bug in default allocation
- Fixed bug in ngen schema

## Testing

1. Updated existing tests and added test coverage for new request types.
2. Some tests may fail pending merge of #45 

## Checklist

- [ ] Passes all existing automated tests
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

### Other

- [ ] No linting errors or warnings


stcui007:
I'll review it.

robertbartel (changes requested):
Two things I noticed: 
1. The version number for the _communication_ package needs to be updated 
    - This will probably eventually affect #45 also.
2. I can't remember if there are hidden implications to using the `super(ClassName, self).__init__(...` syntax when the superclass doesn't also use it (_maas_request.py_ line 675).
    - This seems to be tested and fine, but it also seems like `super().__init__` would have worked for this case.
    - I figured I would mention it since it caught my eye, but I don't think it needs to be changed.

stcui007:
Just wonder if you need a line like:
sudo /usr/sbin/sshd -D
at the end of the entrypoint.sh.

hellkite500:
> Two things I noticed:
> 
> 1. The version number for the _communication_ package needs to be updated
>    
>    * This will probably eventually affect #45 also.
> 2. I can't remember if there are hidden implications to using the `super(ClassName, self).__init__(...` syntax when the superclass doesn't also use it (_maas_request.py_ line 675).
>    
>    * This seems to be tested and fine, but it also seems like `super().__init__` would have worked for this case.
>    * I figured I would mention it since it caught my eye, but I don't think it needs to be changed.
@robertbartel 
The `super()` call was copied from the NWMRequest signature.  It looks like this is legacy python 2.7 style usage of `super`.  The *intent* is to ensure the correct base class constructor is used, but I don't think that is *actually* what is happening here.  Looking at [this](https://stackoverflow.com/questions/26927571/multiple-inheritance-in-python3-with-different-signatures) SO post, it would appear that if this does work, it works for entirely the wrong reasons.  I'll take a look and push up a fix for both the NGEN and the NWM constructors.

I'll also bump the version (thanks for the reminder!)

robertbartel (approved):

PR #50, opened 2020-09-30
Explicit listing of required python packages needed to pass testing.

hellkite500:
To better support testing and development environment setup, the requirements.txt needs to be carefully maintained.  To identify the required packages in this PR, the utility script `update_package -d` was used to install all the DMOD packages and their dependencies from the `setup.py` of each module.  `pip` outputs the list of collected and installed packages, which was copied into the `requirements.txt`.  Any duplicates were removed from the original list and left in the section marked as `minimum required to pass testing`.  The rest were commented out as marked as `dev requirements`. 

A fresh virtual environment was created and these dependencies were installed using `update_package.sh -D` to install only the `requirements.txt` packages.  Then `update_package.sh` was used to install DMOD packages into this environment, and finally `run_tests.sh` verified that the environment was able to support all existing tests.


robertbartel:
@hellkite500, I need to think about this a little more, but there are three things I am worried about:

1. With the dev dependencies commented out, is it accomplishing the file's primary purpose?
    - To be clear, I'm think about both what the primary purpose is and whether it's fulfilled.
1. Do things get overly bloated and/or confusing by combining dev and runtime requirements in one file, especially if transitive dependencies are listed (whether some things commented out or not)? 
    - Alternatively, should we adopt a second file; e.g., an additional _requirements_testing.txt_ file?
2. There is a degree of disconnect from the "true" configuration of transitive dependencies by also making them direct dependencies like this. 
    - I'm not sure whether this could be significant in some situations, however rare they may be.
    - Keep in mind that a rare situation that came up and caused a problem is what prompted this issue being created.

hellkite500:
This has gotten a little stale, am going to close this for now and can revist later if needed.

Issue #49, opened 2020-09-30
Unit testing uses incorrect cache of DMOD package.


hellkite500:
https://github.com/NOAA-OWP/DMOD/blob/f8aaea7d4933a9ac1dc014ec26503ee25b75cc7a/.github/workflows/testing_and_validation.yml#L33

The cache key is static, so as long a the cache has been built the same one will be used each time.  But since the DMOD packages are installed into this venv, if multiple changes to various package components are made in subsequent PR's, the testing is likely to fail.

Need to invalidate this venv cache when DMOD packages are changed, or not cache DMOD installs and only cache external dependencies.

hellkite500:
Closed by #51

PR #51, opened 2020-10-01
Make Action venv cache key depend on dependencies.

robertbartel:
Adjusting the cache key for potentially loading the Python virtual environment during Github Action for testing and validation, such that the key is deterministically based on setup.py files for contained Python packages, thereby ensuring the cache is not used if any dependencies changed.

This should address #49.  PR #50 was also intended to help, and while it has other benefits and should continue being reviewed, I think this is a better solution to the immediate problem (assuming it works; can't really test it until something fires the Action off).

### DIFF
--- a/.github/workflows/testing_and_validation.yml
+++ b/.github/workflows/testing_and_validation.yml
@@ -56,9 +56,11 @@ jobs:
           mkdir ssl
           mkdir ssl/local
           mkdir ssl/requestservice
+          mkdir ssl/requests
           mkdir ssl/scheduler
           ./scripts/gen_cert.sh -d ssl/local -email robert.bartel@noaa.gov
           cp -a ssl/local/*.pem ssl/requestservice/.
+          cp -a ssl/local/*.pem ssl/requests/.
           cp -a ssl/local/*.pem ssl/scheduler/.
 
       - name: Run Tests
--- a/.github/workflows/testing_and_validation.yml
+++ b/.github/workflows/testing_and_validation.yml
@@ -25,12 +25,15 @@ jobs:
         run: sudo apt-get install -y python3-venv
         timeout-minutes: 5
 
+      - name: Prepare Contextual Dependency Hash Key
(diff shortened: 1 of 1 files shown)


hellkite500 (approved):

hellkite500:
I think this temporarily solves the problem, but I'm not convinced we won't see it again.
If package `A` uses components of package `B` to test functionality of package `A`, then
> `A/setup.py`
will contain a requirement of `B >= x.y.z`.

If package `B` is updated, the requirement in `A` gets bumped, the hash changes, the cache misses, everything works for that instance of the test.

But then what happens if:
1.  The changes in `B` break a test in `A` and require changing?  
2. Or review of `B` by itself simply requires changes/modifications before being accepted.

WIth this fix, potentially every commit that triggers a CI run may need to bump versions.

robertbartel:
@hellkite500 you are correct.  While in theory our processes could keep us out of such situations, realistically that won't happen.  Additionally, I don't really account for a change in _requirements.txt_, which the _update_package.sh_ script also utilizes when building the virtual environment.  

I have another idea in mind, so I need to make a few adjustments to this before we move forward with it.

robertbartel:
Ok, @hellkite500, I think this should work now.   Here is the process:

- The Action caches the Python venv, keyed by a hash of the _\_version.py_ files and requirements file(s)
  - If there was a recent run, and the versions and requirements haven't changed, the cached venv is loaded
- Another cache is created for a directory of package-specific checksum files, itself keyed by a hash of all the files in all directories under _python/_
- Checksum files are generated for all (finished) packages within this cached checksums directory when it is initially created
    - This should coincide with whenever the cache misses, which in turn should coincide with when the above venv directory cache misses, which in turn should mean a new venv was built
- After the checksum directory exists, if a package's generated checksum doesn't match the expected, that particular package is updated using the _update_package.sh_ script
    - Again, this is skipped if the venv cache didn't hit, because that should mean there is a fresh venv

hellkite500 (approved):

PR #55, opened 2020-10-06
Docker updates

hellkite500:
Updates to the build and runtime of model images.

Key changes in the compose build steps better align model version/naming with docker version tag semantics.

The entry points of the model images needed to be in `[]` form along with `CMD []` to allow the entrypoint to be executed when the service is created and the args passed correctly to the entrypoint.

The NWM entrypoint was shored up against some particular issues involving a race condition with the SSH daemon being started vs the ssh service ready to respond to requests on the ssh port.

### DIFF
--- a/docker/main/docker-deploy.yml
+++ b/docker/main/docker-deploy.yml
@@ -10,7 +10,6 @@ services:
         - "3013:3013"
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
-      - ../../${DOCKER_VOL_DOMAINS?}:/nwm/domains
       - ../../ssl:/ssl
       - ./schedulerservice/resources.yaml:/code/resources.yaml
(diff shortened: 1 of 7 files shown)


hellkite500:
Note that the entry points assume some of the structure reflected in `scheduler-service/image_and_domain.yaml`.

docker/main/docker-deploy.yml
```diff
@@ -10,7 +10,6 @@ services:
         - "3013:3013"
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
-      - ../../${DOCKER_VOL_DOMAINS?}:/nwm/domains
```
robertbartel, line 13: @hellkite500, don't we need this to pull the domain data into the environment at runtime?

hellkite500: This is mounted in to the individual workers and configured via the `load_image_and_mounts` in the `Launcher` class of `scheduler.py`

PR #69, opened 2021-04-06
Minor updates

hellkite500:
Couple minor fixes I had to apply to get get the full stack up and running.  Not sure where the python versions diverged, but the `djangorestframework` missing dependency fell through in #65.

We probably wanna get some testing up for django/gui components in the near future if this is going to be the official path forward.


christophertubbs (approved):
Looks like this is exactly what we need

christophertubbs (approved):

PR #86, opened 2021-06-29
Generalizing Design of External MaaS Request Types and Clients

robertbartel:
Previously the _MaaSRequest_ (and subtypes, associated response types, client types, etc.) was the most general version of a request expected to be generated externally (i.e., to be directly generated by a user-initiated event).  However, it also was designed to implicitly represent a request for a model execution to be performed.

In general, other types of MaaS request could (and very shortly, _will_) exist; e.g., partitioning config generation requests.  This change introduces another layer of abstraction at various points to allow better generalization of _MaaSRequest_, etc.


hellkite500:
I need to do a deeper inspection of the changes, but one thing that came to mind glancing at them is that while we are touching this part of the code, could/should we refactor all the `MaaS/maas` names to `dmod` for consistency?

robertbartel:
@hellkite500 that's a fair question, though it may be hard to answer.   E.g., there are still a lot of references to "model as a service" in NextGen docs and discussions.  

Either way, I think something of that nature will probably need its own task/PR.

python/lib/communication/dmod/communication/maas_request.py
```diff
@@ -484,10 +494,12 @@ def to_dict(self) -> dict:
 
 
 class MaaSRequestResponse(Response, ABC):
-
     def __init__(self, success: bool, reason: str, message: str = '', data=None):
         super().__init__(success=success, reason=reason, message=message, data=data)
 
+
+class ModelExecRequestResponse(MaaSRequestResponse, ABC):
```
hellkite500, line 501: Why does this subclass ABC?

hellkite500 (approved):

python/lib/communication/dmod/communication/maas_request.py
robertbartel: In my mind, the `ModelExec`-layer needed to be added to differentiate between general user-initiated requests (i.e., the `MaaS`-layer request/response) and user-initiated requests specifically for running a model.  To be consistent, I thought both the request and the response subtypes should be included.

But even with this new separation, these are both still abstractions.   Only some subtype of `ModelExecRequestResponse` should ever be directly instantiated, or there would be some ambiguity (or at least inconsistency with the other request/response usages).   So I added the explicit `ABC` subclassing to reenforce the intended design.

PR #90, opened 2021-07-06
add option for deploying minio object store

hellkite500:
An object storage solution will aid in the development of moving configurations from user -> model, and data from model -> user.  This PR adds a configurable [minio](https://docs.min.io/docs/) docker deployment to the stack.  It can be spun up with `scripts/control_stack.sh object_store deploy`.

## Additions

- object store

## Todos

- A future task will be automating service access keys for various components of the stack, along with appropriate access policies.

## Checklist

- [x] Passes all existing automated tests
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

### Target Environment support

- [x] Docker


docker/object_store/docker-compose.yml
```diff
@@ -0,0 +1,116 @@
...
+      - mpi-net
+      - requests-net
+    deploy:
+      restart_policy:
+        delay: 10s
+        max_attempts: 10
+        window: 60s
+      placement:
+        constraints:
+          - node.labels.minio1==true
```
robertbartel, line 27: On first glance, my thought for this was we need to either automate adding these labels or at least automate a check for them prior to this stage of starting the stack.  Thinking more, that is less than ideal because of the ambiguity of how any arbitrary swarm might be set up, and exactly where nodes might need to be placed. But even so, it seems to me that if we are defining each instance with its own service - effectively fixing the size of the minio cluster - it may still be best to implement direct control over the node setup/placement. 

That said, I'm wondering if we shouldn't try a different approach. Instead of multiple services, just have one service definition, but have it set up with replicas. This would also seem to make the minio cluster size easily configurable per environment. Swarm node labels could still be used locally for controlling placement, and it would be more sensible to require this to be set up manually if it enabled greater flexibility and control. 

It's possible you already considered all this and the constrains of swarm or minio prevent it somehow. Otherwise, might be worth talking through in a bit more detail.

robertbartel (commented):
@hellkite500 I'm curious about replicas in the stack as an alternative. See my in-line comment.

docker/object_store/docker-compose.yml
hellkite500: The problem is with the underlying storage.  If you had a distributed file system where some other service guaranteed replication of data across the hosts, then you could do this.  Docker swarm does not replicate volumes (they are local to each node) so you have to pin the service to the host its volume is associated with.

PR #94, opened 2021-08-26
bump nginx versions to 1.21

hellkite500:
Bump nginx docker image versions.


christophertubbs (approved):
Pretty reasonable - both images should probably be the same just for consistency, but it's a safe course of action for now.

PR #110, opened 2022-01-11
Adjust GUI requirement for requests package to fix urllib3 issue

### LINKED ISSUE #105: Ensure compatibility with urllib3 1.26.5
As described in #101, there appear to be some problems in at least some environments with the current required version of urllib3.  These need to be investigated and addressed.

The move to this version was done via an automated workflow (#80), so reverting to a previous version should not be done without careful examination.

robertbartel:
Loosening restriction of specific version for requests package as part
of the requirements for the GUI Python code, as the previous version was
causing a dependency conflict after the move to urllib3 1.26.5 (closes
#105).

### DIFF
--- a/python/gui/requirements.txt
+++ b/python/gui/requirements.txt
@@ -7,7 +7,7 @@ idna==2.8
 jsonschema==3.0.2
 pyrsistent==0.15.4
 pytz==2019.2
-requests==2.22.0
+requests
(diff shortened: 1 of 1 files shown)


christophertubbs (approved):
That's a good approach - unless you're doing something really out there, just about any version of requests from the last several years works.

PR #113, opened 2022-01-14
Update ngen and ngen-deps Docker images

### LINKED ISSUE #100: NGen Docker Image Configuration Outdated
The main Docker image for running Nextgen has become out of date.  It needs to be reviewed and made current.  It may be worth considering whether it should be expanded in some ways also (e.g., building separate images for parallel versus serial framework executables).

One specific observation has been that an error occurs during image builds at the step when the framework is compiled.  The Docker image config is not installing a supported Python version, but the lack of Python is not accounted for properly when the image build compiles the framework.

robertbartel:
Updating the `ngen` and `ngen-deps` Docker images to be up to date with the current revision of the Nextgen framework code.

This will close #100.


christophertubbs (approved):
Looks solid and very much needed to me.

PR #130, opened 2022-01-31
Fixes and improvements to Docker stack image builds and dependencies

robertbartel:
Fixes and improvements to Docker stack image builds and dependencies.

- Add Python package version constraint to `schedulerservice` and `requestservice` Docker images
- Optimizing `py-sources` Docker image build layers for fewer long rebuilds
- Default to better versions for NetCDF packags in `nwm-deps` image
- Bump model execution `base` image to be based on alpine 3.15
- Fix `base` Docker image apk package add command
- Add `pandas` to `ngen-deps` image apk packages
- Fix `nwm-deps` image build for `apk` '--repository' use
- Fix MPI build issue in `nwm-deps` in new alpine
- Fix bug in `nwm-deps` with MPI build and job count

PR #131, opened 2022-01-31
Bug fixes for GUI and Websocket connections

robertbartel:
- Fix typo in `nwm_gui` Docker stack `app_server` service volume config
- Fix bug with WebSocketInterface related to extending MaaSRequest class to ModelExecRequest 
- Bumping Python package deps and versions
- Updating version of `dmod.communication` package required by GUI code


mattw-nws:
Am I missing it or are all the dependency bumps OWP software? No external dependencies actually changed?

robertbartel:
This PR probably needs to be merged prior to final review of #130.

robertbartel:
@mattw-nws, correct.  These are internal dependency relationships between the different namespace packages.  There is just a bit of a domino effect.

As a brief aside: the package architecture probably needs to be revisited to avoid the dependency complexity we are seeing right now, but that will have to be addressed later.

python/services/requestservice/setup.py
```diff
@@ -14,7 +14,7 @@
     author_email='',
     url='',
     license='',
-    install_requires=['websockets', 'dmod-communication>=0.4.1', 'dmod-access>=0.1.1',
+    install_requires=['websockets', 'dmod-communication>=0.4.2', 'dmod-access>=0.2.0',
```
mattw-nws, line 17: Notably [/python/lib/access/dmod/access/_version.py](https://github.com/robertbartel/DMOD/blob/stack_fixes/websockservice/python/lib/access/dmod/access/_version.py) still shows `0.1.1`, so is this an error or are there missing changes in dmod-access, perhaps?

mattw-nws (changes requested):

mattw-nws (approved):

PR #134, opened 2022-02-07
Some issues found with and revealed by unit tests

robertbartel:
- Fix issue with test helper script that obscured tests failures when run automatically (especially in Github Actions)
- Fix issue with tests for _dmod.communication_ scheduler client, where these would fail if run in the same process as other tests that use the `asyncio` event loop, but otherwise pass just fine
- Fixing another bug, this time in the _dmod.scheduler_ `JobImpl` type, related to extending `ModelExecRequest` from `MaaSRequest` and not updating all usage as needed.


python/lib/communication/dmod/test/test_scheduler_client.py
```diff
@@ -88,15 +88,19 @@ def disable_logging(cls, level: Optional[int] = logging.ERROR):
         logging.disable(level)
 
     def setUp(self) -> None:
-        self.loop = asyncio.get_event_loop()
+        try:
```
hellkite500, line 91: If we are using python 3.7 or newer, we can use [get_running_loop](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop) here.

Also, should probably except just the `RuntimeError`?

hellkite500 (approved):
Might consider adjusting the async try/catch slightly, but this will work fine the way it is for a while at least.  Will note that `get_event_loop` is tagged for deprecation in the future of asyncio.

python/lib/communication/dmod/test/test_scheduler_client.py
robertbartel: @hellkite500, yeah, that seems pretty sensible and easy to do.  I've push a change.

robertbartel:
@hellkite500, @mattw-nws, @christophertubbs: re-requesting review after resolving rebasing conflicts.

hellkite500 (approved):

Issue #128, opened 2022-01-28
Complete initial internal storage infrastructure and workflows


robertbartel:
- [x] #142 
- [x] #144 
- [x] #145
- [x] #147 
- [x] #148
- [x] #152

PR #142, opened 2022-02-16
Initial data flow and data service pre-work

robertbartel:
Adjustments to jobs, job status and execution state classes, and job services, in preparation to include the necessary pieces for getting forcing, observation, hydrofabric, etc. data in and out of where it needs to be.

The biggest pieces are:

- a new _DataRequirement_ type (plus supporting metadata types) for representing data that something (i.e., a _Job_) needs
- new _JobExecStep_ values to represent points in the job workflow for:
  - the state when a check is performed on whether it is possible to provide a job with data it needs
  - the state when any transformation, obtaining, or filtering of available data is done to get required data
- a redesign of _JobStatus_ to be its own composite type, rather than an enum
- incorporating these things into _Job_ and job manager classes

### DIFF
--- a/python/lib/modeldata/dmod/modeldata/data/__init__.py
+++ b/python/lib/modeldata/dmod/modeldata/data/__init__.py
@@ -1,3 +1,4 @@
 from .catchment_data import CatchmentData
 from .data_subset import AbstractDataSubset
 from .forcing_data_handler import ForcingDataHandler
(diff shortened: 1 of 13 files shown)


robertbartel:
This won't pass all test until the changes in #134 are merged in (ironically, it probably will also fail silently, and thus actually pass the checks).

christophertubbs (approved):
Looks solid

robertbartel:
Relates to #128.

PR #143, opened 2022-02-16
Communication package improvements

robertbartel:
Improvements to the _dmod.communication_ package, needed in preparation for other service additions and changes.


robertbartel:
Note that some tests may not pass until the resolution of #134.

hellkite500 (approved):

PR #147, opened 2022-03-02
Updates to Object Store Service Config and Helper Scripts

robertbartel:
Changes to the Docker service configuration and helper scripts needed for the object store service stack, in particular when working in a Docker Desktop environment.

Also documentation on initial setup for object store stack.


robertbartel:
The PR for this contains everything in #145 as well, which is why it is initially being created as a draft.  It will be considerably smaller when it is actually ready for formal review (I think only those changes on and after 2/28).

PR #148, opened 2022-03-04
Integrating Dataset functionality with job scheduling and Docker image for ngen worker container

robertbartel:
Work related to preparing the Docker image itself to handle data via _Datasets_, in particular object store datasets that have the backing data bucket mounted directly into the container file system.  Also updating scheduler code to integrate this functionality.

Note that this should remain a draft PR until #147 is complete.


python/lib/scheduler/dmod/scheduler/scheduler.py
```diff
@@ -237,6 +268,169 @@ def build_host_list(job: 'Job') -> str:
...
+        https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact
+        """
+        # TODO (later): handle non-model-exec jobs in the future
+        if job.model_request.event_type != MessageEventType.MODEL_EXEC_REQUEST:
+            raise RuntimeError("Unsupported requested job event type {}; cannot generate Docker CMD arg values".format(
+                job.model_request.get_message_event_type()))
+
+        # TODO (later): have something more intelligent than class type to determine right entrypoint format and
+        #  values, but for now assume/require a "standard" image
+        if isinstance(job.model_request, NWMRequest) or isinstance(job.model_request, NGENRequest):
```
hellkite500, line 347: should this be `not isinstance(....) or not isinstance(...)`?

python/lib/scheduler/dmod/scheduler/scheduler.py
```diff
@@ -364,6 +551,17 @@ def start_job(self, job: 'Job') -> Tuple[bool, tuple]:
...
         service_per_allocation = []
 
+        # Get the Docker Secrets for object store data access
+        # TODO (later): might need to expand to use different users for different situations (create JobType?)
+        secrets = [self.docker_client.secrets.get('object_store_exec_user_name'),
+                   self.docker_client.secrets.get('object_store_exec_user_passwd')]
+
+        # Add this to env variables, depending on whether this is a single node (Desktop) or multi-node setup
+        if len(self.docker_client.nodes.list()) == 1:
+            env_vars = {'MINIO_DEPLOYMENT': 'standalone'}
```
hellkite500, line 561: Once again am wondering why the proxy isn't used even on a single host???

docker/main/ngen/entrypoint.sh
```diff
@@ -2,26 +2,126 @@
...
+S3FS_PASSWD_FILE="${HOME}/.passwd-s3fs"
+
+build_non_s3_object_store_url()
+{
+    # Dataset category (lower case) is $1
+    # For now, have all forcings come from minio1 and other stuff work on minio2
+    # TODO (later): do something a little more intelligent
+    _OBJ_STORE_MINIO_HOST_SELECT=""
+    # Account for local development environment maybe running a single node standalone MinIO deployment
+    if [ "${MINIO_DEPLOYMENT:-}" == "standalone" ]; then
```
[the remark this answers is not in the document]
robertbartel: Yeah, I don't remember why I thought it was necessary, but I was trying to control which minio service handled which type of data.  Which does seem exactly against the point of the proxy.  I've taken the URL selection function out.

But, as you noticed below, it didn't really even matter ...

python/lib/scheduler/dmod/scheduler/scheduler.py
robertbartel: Yes, thanks.  Fixed.

python/lib/scheduler/dmod/scheduler/scheduler.py
robertbartel: I don't think I intended to have this for anything other than the minio setup, and searching the branch now, I don't find any usage of this string elsewhere.  I'll take it out also.

hellkite500 (approved):

PR #150, opened 2022-03-16
Added missing requirements for scikit-learn

christophertubbs:
Fixing the broken unit tests due to the lack of scikit-learn


hellkite500:
Great, now the tests don't even run 😆

christophertubbs:
Yeah... I don't know what's up with tests not running...

It should be fixed, though. The initial fix was adding the requirement, then some logic had to be tweaked so that unscalable values aren't scaled.

robertbartel (approved):

PR #154, opened 2022-03-18
Creation of dmod.dataservice package and Docker service

### LINKED ISSUE #138: Create Python data service package and Docker service

robertbartel:
- New _dmod.dataservice_ service Python package
- New Docker image configuration and entrypoint script for _data-service_  service
- Addition of _data-service_ service in build and deployment configs of _main_ Docker stack
- A few other supporting changes

This PR closes #138.

This PR should remain in draft status until #152 is complete.


docker/main/docker-deploy.yml
```diff
@@ -110,6 +110,43 @@ services:
...
     entrypoint: ["python3", "-m", "${PYTHON_PACKAGE_NAME_SUBSET_SERVICE:?}"]
     command: ["--files-directory", "/hydrofabric_data", "--port", "${DOCKER_SUBSET_CONTAINER_PORT:-5000}"]
+  data-service:
+    image: ${DOCKER_INTERNAL_REGISTRY:?}/dmod-data-service
+    networks:
+      - requests-net
+    deploy:
+      #mode: global
+      #placement:
+      #  constraints:
```
hellkite500, line 120: Any reason there is NOT a constraint on this?  Should we create a label `data-node` or something we can pin these to?  Seems like we wouldn't want this on the compute nodes (though we should probably create a `compute-node` label for those as well if we are going that route.

python/services/dataservice/dmod/dataservice/__main__.py
```diff
@@ -0,0 +1,96 @@
...
+from . import name as package_name
+from .service import ServiceManager
+from pathlib import Path
+from socket import gethostname
+
+
+def _handle_args():
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('--host',
+                        help='Set the appropriate listening host name or address value (NOTE: must match SSL cert)',
```
hellkite500, line 11: Anything more than a couple args makes we wonder if we shouldn't move to a config file (yaml?)...

docker/main/docker-deploy.yml
robertbartel: I think I deemed a constraint non-essential at the time, and I didn't want to worry about having to properly create and document additional _.env_ variables.  I suppose we could just change these to say `DOCKER_DATASERVICE_DEPLOY_CONSTRAINT_1` and `2` instead of `REQUESTS`, rely on variable-replacement defaults, and just not document them fully yet.  But I'm not sure I wouldn't rather leave that until a time when, as you suggest, we properly refine exactly what constraint configuration we need, here and elsewhere.

python/services/dataservice/dmod/dataservice/__main__.py
robertbartel: Possibly, but that will also require either:
1. a mechanism for generating such files dynamically when starting stacks, from the configured project-level environment variables, or 
2. that users be responsible for edit multiple config files, likely with there being some things edited in more than one place that then manually need to be kept in sync

The project-level configuration mechanisms are in need of a significant overhaul, at which point something like _1._ would be easier.  For now though, I don't immediately see another way of keeping a single source of config settings.

docker/main/docker-deploy.yml
robertbartel: I went ahead and opened #164 for this, but let me know if you think we need to do something more immediately.

hellkite500: That should suffice to track this and develop a more comprehensive strategy.   Worst case scenario, this can be manually tweaked to support a particular deployment for the time being.

hellkite500 (approved):

PR #157, opened 2022-03-29
Addition of Partitioning DMOD service

robertbartel:
Addition of service package, service image, worker image, and other changes for partitioning functionality.

This PR should remain in draft status until #156 is complete.


robertbartel:
Note that while checks fail for this, they pass once again in PR #158, which extends directly from the last commit in this PR branch.  As such, there probably is no need to fix an issue here.

python/services/partitionerservice/dmod/partitionerservice/service.py
```diff
@@ -0,0 +1,386 @@
...
+                await websocket.send(str(response))
+
+        except TypeError as te:
+            logging.error("Problem with object types when processing received message", te)
+        except websockets.exceptions.ConnectionClosed:
+            logging.info("Connection Closed at Consumer")
+        except asyncio.CancelledError:
+            logging.info("Cancelling listener task")
+
+    async def manage_job_partitioning(self):
```
hellkite500, line 337: Curious to see where this gets used...

requirements.txt
```diff
@@ -23,4 +23,5 @@ hydrotools.nwis-client
 numpy>=1.20.1
 scikit-learn
 minio
-aiohttp==3.8.1
```
hellkite500, line 26: Any particular reason for the explicit downgrade?

hellkite500 (approved):
Everything looks pretty reasonable here. Left some comments to address and some discussion that could be had, but again these should be able to be addressed post merge if we need to.

PR #158, opened 2022-03-29
Updates to execution workflow for partitioning

robertbartel:
- Addition of status steps to for place in job exec workflow when partitioning should occur or has failed
- Updates to Job and JobManager classes to account for partitioning
- Updates to RequestService for partitioning

This PR should remain in draft status until #157 is complete.


docker/main/ngen-partitioner/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+ARG MPI_ACTIVE=ON
+# Note that, at least for now, these are off by default until the functionality is both generally available and needed
+ARG BMI_FORTRAN_ACTIVE=OFF
+ARG ROUTING_ACTIVE=OFF
+ARG PARTITIONER_EXECUTABLE
+
+RUN git clone --single-branch --branch $BRANCH $REPO_URL \
+    && cd ./ngen \
+    && if [ "x$COMMIT" != "x" ]; then git checkout $COMMIT; fi \
+    && git submodule update --init \
```
mattw-nws, line 27: This may be something you're already working on, but this causes the CFE and TOPMODEL repos to be cloned a second time...which is a surprisingly large amount of bandwidth and storage...maybe me running on my LTE connection is the only person that will ever care, but it definitely makes a difference there.

python/lib/scheduler/dmod/scheduler/job/job.py
```diff
@@ -1099,6 +1151,20 @@ def data_requirements(self, data_requirements: List[DataRequirement]):
         self._data_requirements = data_requirements
         self._reset_last_updated()
 
+    @property
+    def is_partitionable(self) -> bool:
+        """
+        Whether the requirements of the job support partitioning.
+
+        At this time, only NextGen model jobs can be partitioned.
```
hellkite500, line 1159: Would it make sense to try to tie certain job steps and state to a "model" extension and encapsulate them separately from the core scheduling attributes and job steps?  Maybe a job `pre` and `post` step that can a job extension can implement?

python/services/requestservice/dmod/requestservice/service.py
```diff
@@ -99,15 +118,18 @@ async def listener(self, websocket: WebSocketServerProtocol, path):
                             str(result)))
                     await websocket.send(str(response))
                 elif event_type == MessageEventType.MODEL_EXEC_REQUEST:
-                    response = await self._dmod_request_handler.handle_request(request=req_message)
+                    response = self._model_exec_request_handler.handle_request(requests=req_message)
```
hellkite500, line 121: just double checking this `handle_request` is no longer async?

hellkite500 (approved):

python/lib/scheduler/dmod/scheduler/job/job.py
robertbartel: It's an interesting thought.  

For this specifically, I believe the idea of a partitioning step does generalize to abstract job execution, even if the exact details for that step vary.  For a particular job (type) to not require or support partitioning is just a special case.  But dividing up the work for a job is so fundamental to one of the primary motives of DMOD (enough to get its own service created) that I think it belongs.

We will need to adjust the overall design for jobs and job execution, though.   I don't expect those to completely generalize as-is to evaluation and calibration workflows, and certainly not to the eventual workflows that will be desired beyond that.

PR #161, opened 2022-04-02
Fixes to services' Python and infrastructure code

robertbartel:
- Updates to main routine logic and argument handling for several service packages
- Changes to let MRO work properly for init for some multi-inheritance types
- Adding and modifying Python package dependencies where missing or causing dependency issues
- Adjustments to services' deployment configs and Docker images to get services up and running properly

This PR should remain in draft status until #160 is complete.


docker/main/partitionerservice/Dockerfile
```diff
@@ -8,18 +8,34 @@ ARG partitioner_service_package_name
...
+RUN pip install --upgrade pip \
+    && pip install --no-cache --no-index --find-links=/DIST ${partitioner_service_package_name} \
     # After eventually installing all custom packages like this, clean up ... \
     && rm -r /DIST
 
 EXPOSE ${partitioner_service_listener_port}
 
-ENTRYPOINT [ "python3", "-m", "${partitioner_service_module_name}" ]
\ No newline at end of file
+RUN echo "#!/bin/sh" > /helper.sh \
```
hellkite500, line 26: what is this helper for?

hellkite500 (approved):

docker/main/dataservice/Dockerfile
```diff
@@ -8,11 +8,22 @@ ARG service_package_name
...
 RUN apk update \
-    && apk add --no-cache openssl git g++ libffi-dev \
+    && apk add --no-cache openssl git g++ libffi-dev proj proj-dev proj-util gdal-dev openblas openblas-dev \
     && chmod u+x entrypoint.sh \
     && pip install --no-cache-dir --upgrade pip \
-    && pip install --no-cache-dir --find-links=/DIST "${service_package_name}${service_package_constraint}" \
+    && pip install --no-cache-dir aiohttp==3.8.1 \
+    && pip install --no-cache-dir --no-index --find-links=/DIST "${service_package_name}${service_package_constraint}" \
+    # TODO: this is a workaround, and may have side effects (transitive dependency issue with aiohttp and redis) \
+    && pip install --no-cache-dir aiohttp==3.8.1 \
```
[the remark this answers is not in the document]
robertbartel: I can't remember exactly at this point, but I think so.  This may be three times, actually, with the first being redundant, because of transitive dependency constraints when installing the service package.

docker/main/partitionerservice/Dockerfile
robertbartel: I think I added it for debugging purposes at one point.  Probably should be cleaned up, but I'll leave it for now to avoid the entropy at this point.

PR #166, opened 2022-05-23
Evaluations

christophertubbs:
New library used for interpreting passed instructions, reading data, passing them through the metrics library, and yielding a result.


python/lib/evaluations/dmod/evaluations/threshold/retriever.py
```diff
@@ -0,0 +1,26 @@
...
+import typing
+import abc
+
+import pandas
+
+from .. import specification
+from .. import backends
+
+
+class ThresholdRetriever(abc.ABC):
```
robertbartel, line 11: _ThresholdRetriever_, _DataRetriever_, and _CrosswalkRetriever_ appear to basically be the same abstraction.  There are some class methods that appear in some, but I'm not sure there are really needed at the abstract level (even then, for two of the three, there is overlap).   Might be worth refactoring both the abstractions and/or the packaging layout.

python/lib/evaluations/dmod/evaluations/threshold/disk.py
```diff
@@ -0,0 +1,283 @@
...
+from .. import util
+
+from . import retriever
+
+
+def get_datasource(datasource_definition: specification.DataSourceSpecification) -> retriever.ThresholdRetriever:
+    return __FORMAT_MAPPING[datasource_definition.backend.format](datasource_definition)
+
+
+class JSONThresholdRetriever(retriever.ThresholdRetriever):
```
robertbartel, line 24: While more clear in the abstractions (see comment below), it feels like there is a lot of overlap in the _JSONDataRetriever, _JSONThresholdRetriever_, and _JSONRetriever_ (for crosswalk) implementations.  It would be nice to abstract some parts; e.g., perhaps to a mixin or something.   But this isn't necessary and could certainly be deferred until later.

python/lib/evaluations/dmod/evaluations/transformation.py
```diff
@@ -0,0 +1,72 @@
...
+            initial_frame: pandas.DataFrame,
+            observation_key: str,
+            prediction_key: str = None,
+            *args,
+            **kwargs
+    ):
+        ...
+
+
+class ReindexTransformation(TransformationFunction):
```
robertbartel, line 28: I may have missed it, but I don't see this actually being used anywhere (or the superclass above, except for here) ...

python/lib/metrics/dmod/test/test_metadata.py
```diff
@@ -17,7 +17,7 @@ def test_metadata_retrieval(self):
         """
         found_metrics = metrics.get_all_metrics()
 
-        self.assertEqual(len(found_metrics), 12)
+        self.assertEqual(len(found_metrics), 13)
```
robertbartel, line 20: I see this test fail, seemingly expecting 14.

robertbartel (changes requested):
There is at least one essential change needed, as I have a unit test failing when I run it locally (see comment below).  

There is also another comment about some types that seem to have been added but never used, so that should be confirmed.

I added a few more comments as well for things to consider that might be worth doing, but they aren't essential.  

Additionally, things I'll only mention here that aren't critical:
- Some additional docstring would be useful in a few places, though I think that's already been discussed
- Some documentation in the README would be helpful
- These packages are very much self contained compared to the other DMOD Python packages.  That's fine now that they are done, but it may be worth having a more in depth discussion about whether we want to keep things that way.

python/lib/evaluations/dmod/evaluations/transformation.py
christophertubbs: Turns out the `TransformationFunction` stuff ended up getting replaced by the `ApplicationRules` in the data model - I can just yank this out. Good find.

python/lib/metrics/dmod/test/test_metadata.py
christophertubbs: It wasn't accounting for `ProbabilityOfFalseDetection` and the new `LinearTemporalTrendAbsoluteError`. I'm fixing the number and adding the missing checks.

python/lib/evaluations/dmod/evaluations/threshold/retriever.py
christophertubbs: While `DataRetriever` (the class getting evaluation input) and `CrosswalkRetriever` (the class getting/interpreting the crosswalk) have their own little quirks/assumptions, `ThresholdRetriever` really is a good intersection between all of them. I'm gonna see what I can do to distill all of that while I'm looking at it. Good find.

python/lib/evaluations/dmod/evaluations/threshold/disk.py
christophertubbs: I'm probably going to rename `JSONRetriever` to `JSONCrosswalkRetriever`, but I think most if not all of the overlap will be resolved via the distillation of `ThresholdRetriever` as a root class. Those classes are pretty different outside those common functions.

python/lib/evaluations/dmod/evaluations/specification/model.py
```diff
@@ -0,0 +1,2020 @@
...
+        converted_values = [
+            convert_value(member, expected_type[0])
+            for member in value
+        ]
+        return converted_values
+
+    return value
+
+
+def create_class(cls, data, decoder: json.JSONDecoder = None):
```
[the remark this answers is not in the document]
christophertubbs: I went ahead with `create_class_instance`.

christophertubbs:
I addressed the requested changes. Two sets of unit tests are still failing, and those are `communication` and `modeldata`, so pretty disconnected from this domain.

christophertubbs:
Alright, I ended up resetting and recommitting changes. The tests failed, but those are, again, the usual suspects.

robertbartel (approved):

PR #172, opened 2022-07-14
Client Python package and basic CLI client 

robertbartel:
- Adding the new _dmod.client_ library package
- Adding classes for client-side behavior for interacting with services for job requests and dataset management
- Adding a CLI client via _dmod.client_ for running DMOD actions , including listing datasets
- Adjusting and improving the design of several message subtypes, in particular with `*args` and `**kwargs` for better/corrected multiple inheritance support


python/lib/client/dmod/client/__main__.py
```diff
@@ -0,0 +1,206 @@
...
+        elif args.command == 'dataset':
+            execute_dataset_command(args, client)
+        else:
+            raise ValueError("Unsupported command {}".format(args.command))
+
+    except Exception as error:
+        print("ERROR: {}".format(error))
+        exit(1)
+
+    blah = ''
```
hellkite500, line 202: `blah`....cause why not?

python/lib/client/dmod/client/__main__.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+if __name__ == '__main__':
+    main()
```
hellkite500, line 206: posix newline?

python/lib/client/dmod/client/client_config.py
```diff
@@ -0,0 +1,119 @@
...
+        if dir_path.is_dir():
+            return dir_path
+        else:
+            raise RuntimeError("Non-existing {} SSL directory configured ({})".format(service_key, dir_path))
+
+    def __init__(self, client_config_file: Path, *args, **kwargs):
+        super(ClientConfig, self).__init__(*args, **kwargs)
+        self._config_file = client_config_file
+        with self._config_file.open() as file:
+            self._backing_config = yaml.safe_load(file)
```
hellkite500, line 65: it is probably worth a comment here the the class's properties depend on the loading/state of `_backing_config` so they must not be accessed until after the config is loaded

python/lib/client/dmod/client/client_config.py
```diff
@@ -0,0 +1,119 @@
...
+        self._dataservice_endpoint_uri = None
+        self._dataservice_ssl_dir = None
+
+    @property
+    def config_file(self) -> Path:
+        return self._config_file
+
+    @property
+    def dataservice_endpoint_uri(self) -> Optional[str]:
+        if self._dataservice_endpoint_uri is None and self.dataservice_hostname is not None:
```
hellkite500, line 78: why check if hostname is not None, but not the port?

python/lib/client/dmod/client/dmod_client.py
```diff
@@ -0,0 +1,83 @@
...
+    @property
+    def requests_endpoint_uri(self) -> str:
+        return self.client_config.requests_endpoint_uri
+
+    @property
+    def requests_ssl_dir(self) -> Path:
+        return self.client_config.requests_ssl_dir
+
+    def print_config(self):
+        print(self.client_config.config_file.read_text())
```
hellkite500, line 61: why not just `self.client.print_config`?

python/lib/client/dmod/client/dmod_client.py
```diff
@@ -0,0 +1,83 @@
...
+            List of one or more paths of files to upload or directories containing files to upload.
+
+        Returns
+        -------
+        bool
+            Whether uploading was successful
+        """
+        return await self.dataset_client.upload_to_dataset(dataset_name, paths)
+
+    def validate_config(self):
```
hellkite500, line 81: Could I interest you in pydantic ;)

python/lib/client/dmod/client/request_clients.py
```diff
@@ -0,0 +1,206 @@
...
+        # TODO: (later) consider also adding param for data to be added
+        request = DatasetManagementMessage(action=ManagementAction.CREATE, dataset_name=name, category=category)
+        self.last_response = await self.async_make_request(request)
+        return self.last_response is not None and self.last_response.success
+
+    async def list_datasets(self, category: Optional[DataCategory] = None) -> List[str]:
+        action = ManagementAction.LIST_ALL if category is None else ManagementAction.SEARCH
+        request = DatasetManagementMessage(action=action, category=category)
+        self.last_response = await self.async_make_request(request)
+        return self._parse_list_of_dataset_names_from_response(self.last_response)
```
hellkite500, line 65: so wouldn't this technically return `None` since the function is "unimplemented" with `pass`?

python/lib/client/dmod/client/request_clients.py
```diff
@@ -0,0 +1,206 @@
...
+        """
+        Attempt to set the session information properties needed to submit a maas request.
+
+        Parameters
+        ----------
+        use_current_values : bool
+            Whether to use currently held attribute values for session details, if already not None (disregarded if
+            ``force_new`` is ``True``).
+        force_new : bool
+            Whether to force acquiring a new session, regardless of data available is available on an existing session.
```
hellkite500, line 107: This sentence (at the end) is not real clear.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -0,0 +1,206 @@
...
+                                               dataset_name=name, category=category)
+        self.last_response = await self.async_make_request(request)
+        return self.last_response is not None and self.last_response.success
+
+    async def list_datasets(self, category: Optional[DataCategory] = None) -> List[str]:
+        await self._async_acquire_session_info()
+        action = ManagementAction.LIST_ALL if category is None else ManagementAction.SEARCH
+        request = MaaSDatasetManagementMessage(session_secret=self.session_secret, action=action, category=category)
+        self.last_response = await self.async_make_request(request)
+        return self._parse_list_of_dataset_names_from_response(self.last_response)
```
hellkite500, line 172: again, does this really return a bool?

python/lib/client/setup.py
```diff
@@ -0,0 +1,24 @@
...
+        long_description = readme.read()
+except:
+    long_description = ''
+
+exec(open('dmod/client/_version.py').read())
+
+setup(
+    name='dmod-client',
+    version=__version__,
+    description='Client interface package for components of the National Water Model as a Service architecture',
```
hellkite500, line 14: "the National Water Model as a Service architecture" -> "the DMOD architecture"?

hellkite500 (dismissed):

python/lib/client/dmod/client/client_config.py
robertbartel: I added a few docstrings for the type and the function.

python/lib/client/dmod/client/client_config.py
robertbartel: Good point.  I will fix.

python/lib/client/dmod/client/dmod_client.py
robertbartel: Doesn't appear to be a good reason.  I'll change that.

python/lib/client/dmod/client/request_clients.py
robertbartel: You are correct, and it would have probably been a better practice to have that function return an empty list.  In the future, I will try to remember this and do something along those lines instead.

This one I am going to leave as-is, in the interest of reducing branch merging complexity.  An implementation will come along in another PR branch before too much longer.

python/lib/client/dmod/client/request_clients.py
robertbartel: For now I'd like to leave this one alone.  I need to clean up some things related to these functions, but will do that in a later branch.

python/lib/client/dmod/client/request_clients.py
robertbartel: It _will_ ... eventually ...

python/lib/client/setup.py
robertbartel: Good catch.  I'll fix.

hellkite500 (approved):

PR #177, opened 2022-08-01
Updated out of date references/comments in the GUI files

christophertubbs:
The original proof of concept for the GUI came from a quick and dirty stab based on some resources from the WRES GUI. As a result, there are a few references to the WRES that are wildly low hanging fruit. I figured it would be easy enough to just go ahead and fix them.

There are also some references to `maas_experiment`. If I remember correctly, that `maas_experiment` really was just a reference to the original PoC being an experiment. That should _probably_ be fixed, but there is a higher chance of bugs considering that it would result in the renaming of directories and resources, not just a few comments or aliases.


hellkite500 (approved):

aaraney:
From the conflicts shown, it looks like this has be resolved. Closing.

PR #179, opened 2022-08-03
Additional client package and related changes

robertbartel:
Several changes to client package and core, in particular in support of dataset client capabilities, NextGen request client capabilities, and general CLI functionality.

This PR should remain in draft status until #178 is complete.

PR #180, opened 2022-08-04
More fixes for data management and data client

robertbartel:
- Various adjustments to `Dataset`, including fixes for serialization and deserialization
- Adjustments to `DatasetManager`
- Adjustments to object store stack/services
- Work to get object store based datasets accessible within workflows
- Adjustments to dataservice functionality to support various operations (query, check if data available, delete, etc.)
- Adjustments to hydrofabric class after format updates
- Adjustments to things using hydrofabric datasets to ensure proper identification of desired hydrofabric

This PR should remain in draft status until #179 is complete.

### DIFF
--- a/python/lib/modeldata/dmod/modeldata/data/dataset.py
+++ b/python/lib/modeldata/dmod/modeldata/data/dataset.py
@@ -594,6 +594,10 @@ class DatasetManager(ABC):
         pass
 
     @abstractmethod
+    def combine_partials_into_composite(self, dataset_name: str, item_name: str, combined_list: List[str]) -> bool:
+        pass
+
+    @abstractmethod
     def create(self, name: str, category: DataCategory, domain: DataDomain, is_read_only: bool,
                initial_data: Optional[str] = None) -> Dataset:
         """
@@ -621,6 +625,23 @@ class DatasetManager(ABC):
         -------
         Dataset
             A newly created dataset instance ready for use.
+        """
+        pass
+
+    @abstractmethod
+    def delete(self, dataset: Dataset, **kwargs) -> bool:
+        """
+        Delete the supplied dataset, as long as it is managed by this manager.
+
+        Parameters
+        ----------
+        dataset
+        kwargs
+
+        Returns
+        -------
+        bool
+            Whether the delete was successful.
         """
         pass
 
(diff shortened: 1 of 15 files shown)


robertbartel:
As stated in #179, I anticipate the checks for this PR will fail.  This should be the last PR in the sequence we are working through with these particular expected failures.

python/lib/modeldata/dmod/modeldata/data/dataset.py
```diff
@@ -577,6 +593,10 @@ def add_data(self, dataset_name: str, dest: str, data: Optional[bytes] = None, s
         """
         pass
 
+    @abstractmethod
+    def combine_partials_into_composite(self, dataset_name: str, item_name: str, combined_list: List[str]) -> bool:
```
christophertubbs, line 597: Not a big deal, but I first thought this had to do with partial python functions, not partial datasets. I wonder if I'm the only one who'd be tripped up or if it might benefit the codebase to come up with a different name like 'combine_partial_datasets' or something. Or I could be even more wrong and it might just need a docstring.

Again, not a showstopper, just something that stuck out to me.

python/lib/modeldata/dmod/modeldata/data/object_store_dataset.py
```diff
@@ -156,32 +157,17 @@ def __init__(self, obj_store_host_str: str, access_key: Optional[str] = None, se
...
-        str
-            The decoded file name that reflects any subdirectory structure.
-
-        See Also
-        -------
-        ::method:`_push_file`
-        """
-        return "/".join(object_name.split(self._OBJECT_NAME_SEPARATOR))
+        try:
+            self._client = Minio(endpoint=obj_store_host_str, access_key=access_key, secret_key=secret_key, secure=False)
```
christophertubbs, line 161: Another non-showstopper that jumped out at me. Yeah, it'll make things more complicated, but I wonder if an external function or dependency injection might help decouple a hard reliance on Minio here. I'm not sure if there are super specific Minio functionality used throughout, so I'm not really qualified to say, but if Minio fulfills the S3 protocol/API/whatever, couldn't other implementations work here? That would allow everyone to use different object stores based on environment. Changing this up later down the line would allow this to be used with Amazon S3, google cloud, CEPH, etc.

Not a showstopper since **IF** it is reasonable to address, it should be a problem for future DMOD considering the work required. There are a ton of other priorities right now.

christophertubbs (approved):
Everything looks good to me and the fact that it serves as a foundational step on one of our deployed environments further cements that.

My only curiosity is whether the failing client check will be fixed via merge or if it'll break the current versions. Either way, it's fairly low risk.

PR #182, opened 2022-08-12
Initial implementation of data transfer functionality

robertbartel:
The PR should remain in draft status until #180 is complete.

PR #185, opened 2022-08-24
Fixes to request-service

robertbartel:
- Fixing some default values and parameter handling in the CLI
- Updating listener to handle new event types correctly
- Adjusting entrypoint script and Docker config 

The PR should remain in draft status until #182 is complete.


docker/main/docker-deploy.yml
```diff
@@ -79,10 +79,16 @@ services:
       - DOCKER_SECRET_REDIS_PASS=myredis_pass
```
christophertubbs, line 79: Comment: It might be worth adding all of these into a .env file or some sort of configuration setup due to the number of variables. Looks like this is going to scale at 3 variables per service? If that's the case a more centralized setup might be helpful in the future (especially since it's all "host", "port", "SSL Directory")

docker/main/requestservice/entrypoint.sh
```diff
@@ -50,4 +50,10 @@ python -m ${SERVICE_PACKAGE_NAME:?} \
     ${_DEBUG_ARG:-} \
     --scheduler-host ${SCHEDULER_ENDPOINT_HOST:?} \
     --scheduler-port ${SCHEDULER_ENDPOINT_PORT:?} \
-    --scheduler-ssl-dir ${SCHEDULER_CLIENT_SSL_DIR:?}
+    --scheduler-ssl-dir ${SCHEDULER_CLIENT_SSL_DIR:?} \
```
christophertubbs, line 53: It might help to add all of these environment variables as possible defaults or 'fall-backs' if they are all required and all read from environment variables. Anything to make it pick-up-and-play. This also feeds into my previous comment of providing a more centralized access point for {service: (port, host, ssl directory)} data.

christophertubbs (approved):
The only things that stood out to me were possible feature additions in the future (see comments for details)

PR #186, opened 2022-08-24
Scheduler and Workflow Executions Fixes

robertbartel:
A wide variety of fixes and improvements to support the initial, successful executing workflow.

The PR should remain in draft status until #185 is complete.

PR #187, opened 2022-08-25
Miscellaneous Cleanup

robertbartel:
A few things that didn't fit especially well into another group of commits:

- Updating entrypoint scripts to make dev environment fast deploy ignore package dependencies
- Changes to several unit tests (both fixes and refactoring)
- Async bug fix in request service

The PR should remain in draft status until #186 is complete.


christophertubbs (approved):
I thought it was weird that some of the model request tests hardcoded json as strings, then read said JSON, where it maybe the manual string could have been an external file that could be read as a string. It leaves a good bit of room to mess up tests based off of basic syntax issues and end up testing the developer's ability to write tests, not necessarily the functionality. Not a showstopper, though. Just weird.

Additionally, renaming functions to be zero indexed made for a good start for the morning. I chuckled.

PR #188, opened 2022-08-25
Changes to Get Multi-Node Swarm Online 

robertbartel:
The final set of changes to get the first alpha version of a multi-node deployment online.  These include:

- Modifications to the object store stack to
  - better distinguish single- and multi-host MinIO deployment configs
  - optimize overall performance, especially through proxy
  - fix a few miscellaneous bugs
- Adjustments/fixes to components for making object store datasets accessible
- Fixes to several Docker image
- Rename of _JobAllocationParadigm_  class to _AllocationParadigm_ and move to _dmod.core_ package
- Adjustments/fixes to several classes and tests related to allocation paradigm
- A few other miscellaneous changes

This PR should remain in draft status until #187 is complete.


python/lib/communication/dmod/communication/scheduler_request.py
```diff
@@ -40,21 +45,22 @@ def factory_init_from_deserialized_json(cls, json_obj: dict):
         try:
             model_request = ModelExecRequest.factory_init_correct_subtype_from_deserialized_json(json_obj['model_request'])
             if model_request is not None:
+                alloc_paradigm = json_obj['allocation_paradigm'] if 'allocation_paradigm' in json_obj else None
```
christophertubbs, line 48: Not enough to derail the PR, but the `'allocation_paradigm'` magic string might need to be defined in a more concrete fashion elsewhere (maybe at the top?). This is probably inconsistent in that I probably haven't noticed this elsewhere and I definitely do it myself. Defining it elsewhere isn't going to prevent the direct usage, but it could help goobers (i.e. me) from typing it in wrong or using a different term elsewhere, like naming it 'allocationparadigm` or something and only finding the error later.

python/services/dataservice/dmod/dataservice/__main__.py
```diff
@@ -38,7 +38,7 @@ def _handle_args():
     parser.add_argument('--object-store-host',
                         help='Set hostname for connection to object store',
                         dest='obj_store_host',
-                        default='minio_proxy')
+                        default='minio-proxy')
```
christophertubbs, line 41: I'm just thinking about this now, but maybe something like checking for an environment variable with a fallback for this value might be helpful. Changing '_' to '-' isn't a huge change, but I've seen it elsewhere. It being a default host (probably a docker container, right?), it might help to have this string be in a central location so that all of these types of variables may be consistent. Plus, overwriting them with environment variables would be easy enough.

Not enough to derail the PR, though. Just an early morning thought.

python/services/evaluationservice/dmod/evaluationservice/service/application_values.py
```diff
@@ -84,7 +84,9 @@ def get_full_localtimezone():
 
     # If this is a system with 'zoneinfo' set up as part of the environment, parse the name from the linked zone
     zone_path = Path("/etc/localtime")
-    zone_path = str(zone_path.readlink()) if zone_path.exists() and zone_path.is_symlink() else ""
+    # This doesn't actually work in Python 3.8, though does in 3.9
+    #zone_path = str(zone_path.readlink()) if zone_path.exists() and zone_path.is_symlink() else ""
+    zone_path = str(os.readlink(str(zone_path))) if zone_path.exists() and zone_path.is_symlink() else ""
```
christophertubbs, line 89: Excellent catch.

python/services/partitionerservice/dmod/partitionerservice/service.py
```diff
@@ -19,7 +19,7 @@
 from uuid import uuid4
 
 logging.basicConfig(
-    level=logging.DEBUG,
+    level=logging.INFO,
```
christophertubbs, line 22: This might need to help boost the priority of moving some logging logic to `dmod.core` or elsewhere so this call doesn't have to be repeated elsewhere. I swear I've seen this exact wording elsewhere. If this one is changed to `INFO`, but another isn't, the level will still be `DEBUG` if that one is hit first (I believe, anyways). If `dmod.partitionerservice.__main__` calls a file that defines `logging.basicConfig`_before_ importing `service`, this doesn't amount to much.

christophertubbs (approved):
Only mild things that don't necessarily fall under the scope of this PR jumped out at me. Those may not be a big issue now, but they might become bigger issues as the code base grows.

python/lib/communication/dmod/communication/maas_request.py
```diff
@@ -257,6 +267,18 @@ def __eq__(self, other):
                     return False
             return True
 
+    @property
```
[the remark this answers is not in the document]
robertbartel: @christophertubbs, not totally sure I understand what you are asking, but I'm going to go with my best guess.  If I'm talking past you, then we can discuss more deeply in a different medium.

`AllocationParadigm` is indeed an enum type, and I think this makes the most sense.  I don't particularly like the idea of putting too much documentation about `AllocationParadigm` here for maintenance and best-practice reasons.  Something like changing the name of the value cited as an example could easily miss having the update applied here.  If there's more that needs to be documented in `AllocationParadigm`, then let me know.

python/lib/communication/dmod/communication/scheduler_request.py
robertbartel: Yeah, I have defined serialization keys as "private" class variables with some-but-not-all of the serializable types.  I agree:  at some point we should go back and refactor that in places it isn't already done.

python/services/dataservice/dmod/dataservice/__main__.py
robertbartel: I have typically tried to use args directly for external parameters.  In my mind, it makes packages less brittle.  

My real mistake may be providing the currently-correct value as a default here and most of the other analogous places for this value, as it is kind of centrally located already.  A DMOD _.env_ variable is set up to be the source of this, which passes through to (parts of) the Docker config, which passes through the entrypoint script, which provides this arg value.   I could take the defaults out and force the DMOD config item to be in place.  But even then, I can't centralize quite everything (e.g., the nginx configs).

python/services/partitionerservice/dmod/partitionerservice/service.py
robertbartel: I think we have to kick that can a little further down the road, but you are correct in that it needs to be reworked.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,5 +1,6 @@
 from abc import ABC, abstractmethod
 from datetime import datetime
+from dmod.core.execution import AllocationParadigm
 from dmod.communication import InternalServiceClient, MaasRequestClient, ManagementAction, ModelExecRequestClient, \
     NGENRequest, NGENRequestResponse
 from dmod.communication.client import R
```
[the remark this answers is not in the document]
robertbartel: I opened #192 to help track this.

PR #189, opened 2022-08-25
Parameterize Dataset Type and Other Fixes

robertbartel:
- Re-adding the use of proxy in single-node MinIO deployment
- Updating/fixing things related to making object store datasets available in worker containers
- Adjustments to _Dataset_ class hierarchy to have there only be that class and make type (e.g. object store dataset) be a property rather than implied via subclassing
- Add support for workaround of performance issues with large (forcing) object store datasets in worker containers via a configurable host bind mount
- Other miscellaneous fixes and improvements

This PR should remain in draft status until #188 is complete.


scripts/init_object_store_dataset_volumes.sh
```diff
@@ -184,6 +184,11 @@ while [ ${#} -gt 0 ]; do
     shift
 done
 
+# Use this as a default
+if [ -z "${S3FS_URL:-}" ]; then
+    S3FS_URL="http://$(hostname):9000/"
```
christophertubbs, line 189: Something should probably be done to externalize the definition for this port. Crossing our fingers and hoping that that `9000` links up across the board may not be reasonable as things grow. I saw a reference to this in python/services/dataservice/dmod/dataservice/service.py. We're looking at some issues if one hardcoded port changes and the other doesn't. I think this hardcoded port of 9000 is in a couple different places, so that leaves a little bit of room to introduce an extremely hard to identify bug.

christophertubbs (approved):

PR #197, opened 2022-10-12
Initial Dataset Management GUI

robertbartel:


python/gui/MaaS/cbv/DatasetApiView.py
```diff
@@ -0,0 +1,62 @@
...
+logger = logging.getLogger("gui_log")
+
+
+class DatasetApiView(AbstractDatasetView):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def _get_dataset_content_details(self, dataset_name: str):
+        result = asyncio.get_event_loop().run_until_complete(self.dataset_client.get_dataset_content_details(name=dataset_name))
```
christophertubbs, line 17: Looks like there are several calls to `asyncio.get_event_loop().run_until_complete`. It might help to have some synchronous versions of these functions on the dataset client since there appears to be some intention to call them synchronously. Maybe.

python/gui/MaaS/cbv/DatasetApiView.py
```diff
@@ -0,0 +1,62 @@
...
+            FileWrapper(custom_filelike, chunk_size),
+            content_type="application/octet-stream"
+        )
+        response['Content-Length'] = asyncio.get_event_loop().run_until_complete(self.dataset_client.get_item_size(dataset_name, item_name))
+        response['Content-Disposition'] = "attachment; filename=%s" % item_name
+        return response
+
+    def get(self, request, *args, **kwargs):
+        request_type = request.GET.get("request_type", None)
+        if request_type == 'download_file':
```
christophertubbs, line 50: This `if` chain leads me to believe that splitting this out across several smaller views might help.

python/gui/MaaS/cbv/DatasetApiView.py
```diff
@@ -0,0 +1,62 @@
...
+            return self._get_download(request)
+        elif request_type == 'datasets':
+            return self._get_datasets_json()
+        elif request_type == 'dataset':
+            return self._get_dataset_json(dataset_name=request.GET.get("name", None))
+        elif request_type == 'contents':
+            return self._get_dataset_content_details(dataset_name=request.GET.get("name", None))
+        if request_type == 'delete':
```
christophertubbs, line 58: This could probably be handled through a `delete` handler instead of a `get` handler. I don't know what all's going on in the workflow (like what the authentication looks like), but it looks like a user could possibly delete datasets through a browser address.

python/gui/MaaS/cbv/DatasetApiView.py
```diff
@@ -0,0 +1,62 @@
...
+            return self._delete_dataset(dataset_name=request.GET.get("name", None))
+
+        # TODO: finish
+        return JsonResponse({}, status=400)
```
christophertubbs, line 62: Django has a specialized response that might help here at `django.http.HttpResponseBadRequest`. If you use it it can help out when diagnosing issues in debug mode.

python/gui/MaaS/cbv/DatasetManagementView.py
```diff
@@ -0,0 +1,114 @@
...
+        kwargs
+
+        Returns
+        -------
+        A rendered page.
+        """
+        errors, warnings, info = extract_log_data(kwargs)
+
+        # Gather map of serialized datasets, keyed by dataset name
+        serial_dataset_map = asyncio.get_event_loop().run_until_complete(self.get_datasets())
```
christophertubbs, line 66: Another `asyncio.get_event_loop().run_until_complete` call. It might help to have two `get_datasets` - `get_datasets` and `get_datasets_async`, with the former just calling the latter with the event loop wrapper.

python/gui/MaaS/cbv/DatasetApiView.py
```diff
@@ -0,0 +1,62 @@
...
+        item_name = request.GET.get("item_name", None)
+        chunk_size = 8192
+
+        custom_filelike = DatasetFileWebsocketFilelike(self.dataset_client, dataset_name, item_name)
+
+        response = StreamingHttpResponse(
+            FileWrapper(custom_filelike, chunk_size),
+            content_type="application/octet-stream"
+        )
+        response['Content-Length'] = asyncio.get_event_loop().run_until_complete(self.dataset_client.get_item_size(dataset_name, item_name))
```
christophertubbs, line 44: `StreamingHttpResponse`s are a little wonky - I don't think they are compatible with the `Content-Length` header and will clash with Gunicorn. It might be better to have a websocket that emits this data and let the client build the blob.

python/gui/MaaS/templates/maas/dataset_management.html
```diff
@@ -0,0 +1,797 @@
...
+                    colCell.style.fontWeight = 'normal';
+                }
+                cellText = document.createTextNode(arguments[i]);
+                colCell.appendChild(cellText);
+                row.appendChild(colCell);
+            }
+            arguments[0].appendChild(row);
+        }
+
+        function buildDatasetFilesEmptyDiv() {
```
christophertubbs, line 154: It might help to define these styles in a css document rather than baking them into the template. They'd be easier to change and help reduce some code bloat.

python/gui/MaaS/templates/maas/dataset_management.html
```diff
@@ -0,0 +1,797 @@
...
+                    let x = 1;
+                },
+                error: function(response) {
+                    alert('Received an error downloading file ' + file_name + ' from dataset ' + dataset_name);
+                    console.log(response);
+                }
+            });
+        }
+
+        function ajaxBuildDatasetFilesTable(dataset_name, contents_json, table) {
```
christophertubbs, line 257: It might be easier to take advantage of a [template tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) here instead of programmatic element creation. That'd help reduce possible nesting errors and make things a little easier to read (since you'd be reading markup instead of js).

python/gui/MaaS/templates/maas/dataset_management.html
```diff
@@ -0,0 +1,797 @@
...
+            $(distributionSelector).removeAttr('disabled');
+            $(distributionRowsSelector).removeClass('disabled');
+            $(distributionRowsSelector).show();
+        }
+        {% endfor %}
+        {% endfor %}
+        */
+    </script>
+    <style>
+        #header {
```
christophertubbs, line 453: We should probably consolidate these and styles from other templates into a common css document to help maintain consistency between pages. Branding, for instance, could be used on just about every page, for example.

python/gui/MaaS/templates/maas/dataset_management.html
```diff
@@ -0,0 +1,797 @@
+<!DOCTYPE html>
```
christophertubbs, line 1: Templates can inherit from base templates and override/fill out different elements from the parent. Using a base template here could probably cut down the size of this template down by a large portion and make it so the only things that need to be written here are the things specific to dataset management and dataset management alone.

python/gui/MaaS/templates/maas/dataset_management.html
```diff
@@ -0,0 +1,797 @@
...
+        }
+
+        #dataset-create {
+            display: none;
+        }
+
+    </style>
+</head>
+<body onload="initGuiViews()">
+<div id="container">
```
christophertubbs, line 590: This sort of boilerplate could be farmed out to a base template so you don't have to worry about all that here.

python/gui/MaaS/cbv/DatasetApiView.py
robertbartel: The base functions for sending data through websockets are (and I am pretty sure must be) `async`, as are the functions making sure the client is authenticated (which relies on the former also).   It would take a decent bit of changing to be able to do this and have everything work, but we can look more at that in the future.

python/gui/MaaS/cbv/DatasetApiView.py
robertbartel: I did originally start to do that, but it occurred to me that this kind of violated the intent of having only a single websocket (at least per authenticated session) open for all client communication.  In fairness, that isn't working properly at the moment, but I think we want to try to have it that way eventually, or else we'd move to REST.  And if we are eventually going for that, I don't think we want too many different views (i.e., because eventually we'll have to consolidate them), although I could be assessing that incorrectly.

python/gui/MaaS/cbv/DatasetManagementView.py
robertbartel: I guess we can, but that seems cluttered.  Not that I haven't already done that kind of cluttering in places ... 

Do you think that would just enhance developer usability of the client class, or were there other advantages you were anticipating?

python/gui/MaaS/cbv/DatasetApiView.py
robertbartel: I don't have this part working anyway, so it can be implemented differently.  I can pull this and other such things out entirely, but I wasn't sure if parts might be useful to show the direction I was trying to go.

python/gui/MaaS/templates/maas/dataset_management.html
robertbartel: Indeed, I started down that road for several things but didn't get everything and wanted to push this.  I plan to do more clean up, it's just a question of whether it's before or after the review and merge of this.

python/gui/MaaS/cbv/DatasetApiView.py
christophertubbs: yeah, if there's a plan to consolidate splitting them out doesn't make a whole lot of sense.

python/gui/MaaS/cbv/DatasetManagementView.py
christophertubbs: I know that I'm the type of goober that would end up calling `self.get_datasets()` and forget to use the asyncio logic. That's just about the only thing that the change would address.

python/gui/MaaS/cbv/DatasetApiView.py
christophertubbs: That's precisely why it should be left in for now. I just wanted to drop some notes while I was looking at it. I don't think any of us has the bandwidth to make major changes at the moment.

robertbartel:
With PRs #197 and #199, I believe once one of them is merged, the other will need modifications to resolve conflicts.

christophertubbs:
I approved #199, so this will need to be modified.

python/gui/MaaS/cbv/AbstractDatasetView.py
```diff
@@ -0,0 +1,29 @@
+from abc import ABC
+from django.views.generic.base import View
+from dmod.client.request_clients import DatasetExternalClient
+import logging
+logger = logging.getLogger("gui_log")
+from .DMODProxy import DMODMixin, GUI_STATIC_SSL_DIR
+from typing import Dict
+
+
+class AbstractDatasetView(View, DMODMixin, ABC):
```
aaraney, line 10: I get the desire to add blanket implementation here for downstream classes, but given that all methods and properties are implemented, is the use of abstract here to indicate to downstream users that they should inherit from this? Still coming up the speed on the codebase so this might be an idiom ive not picked up on yet.

python/gui/MaaS/cbv/DMODProxy.py
```diff
@@ -16,6 +16,8 @@
 from pathlib import Path
 from typing import List, Optional, Tuple, Type
 
+GUI_STATIC_SSL_DIR = Path('/usr/maas_portal/ssl')
```
aaraney, line 19: we might want to mark this as a TODO in the future to make this configurable.

christophertubbs: I think this should also be handled by Django as well. Shouldn't need to be too necessary. Should be in `settings.py`

python/gui/MaaS/cbv/DatasetFileWebsocketFilelike.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def __init__(self, client: DatasetExternalClient, dataset_name: str, file_name: str):
+        self._client = client
+        self._dataset_name = dataset_name
+        self._file_name = file_name
+        self._read_index: int = 0
+
+    def read(self, blksize: int) -> AnyStr:
+
+        result = asyncio.get_event_loop().run_until_complete(
```
aaraney, line 16: do we want to try catch here? if the `download_item_block` call raises an exception and a downstream user of `DatasetFileWebsocketFilelike` catches an exception, the state of their `DatasetFileWebsocketFilelike` could be corrupted or atleast unclear what the behavior should be. Meaning, here, `_read_index`, if caught during a failed read, should continue from the failure point. Is this the expected behavior, or should _read_index be reset to `0`?

python/gui/MaaS/static/maas/js/components/createDatasetForm.js
```diff
@@ -0,0 +1,42 @@
...
+        this.formElementId = this.parentDivId + "-form";
+        this.formContentDivId = this.formElementId + "-div-universal-inputs";
+        this.dynamicVarsDivId = this.formElementId + "-div-dynamic-inputs";
+
+    }
+
+    updateFormatChange(selection) {
+        let dy_div = document.getElementById(this.dynamicVarsDivId);
+        //let initial_length = dy_div.childNodes.length;
+        dy_div.childNodes.forEach(child => child.remove());
```
aaraney, line 13: Related to #198. See [this](https://github.com/NOAA-OWP/DMOD/issues/198#issuecomment-1289327285) comment specifically. This should be changed to the following:

```javascript
// remove all dynamic input child elements
while (dy_div.firstChild) {
  dy_div.removeChild(dy_div.firstChild);
}
```

python/gui/MaaS/cbv/DatasetApiView.py
aaraney, line 17: It seems we might want to wrap these methods (all methods with leading `_`) with try except's and return something meaningful to the client if the request fails. I didn't go looking, but if the result has something like an `ok` property, we might want to return a non-200 here.

python/gui/MaaS/cbv/DatasetManagementView.py
```diff
@@ -0,0 +1,114 @@
...
+        serial_dataset_list = [serial_dataset_map[d] for d in serial_dataset_map]
```
aaraney, line 67: this is nit-picky, but for code readability, `serial_dataset_list = list( serial_dataset_map.values())` might be preferred.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -175,6 +202,50 @@ async def delete_dataset(self, name: str, **kwargs) -> bool:
...
 
+    async def get_dataset_content_details(self, name: str, **kwargs) -> dict:
+        # TODO: later add things like created and last updated perhaps
+        query = DatasetQuery(query_type=QueryType.GET_DATASET_ITEMS)
+        request = DatasetManagementMessage(action=ManagementAction.QUERY, query=query, dataset_name=name)
+        self.last_response: DatasetManagementResponse = await self.async_make_request(request)
+        if self.last_response.success:
+            return self.last_response.data
+        else:
+            return {}
```
aaraney, line 213: should we return something to the caller indicating why, if we know, the response was unsuccessful?

python/lib/client/dmod/client/request_clients.py
```diff
@@ -175,6 +202,50 @@ async def delete_dataset(self, name: str, **kwargs) -> bool:
...
+
+    async def get_item_size(self, dataset_name: str, item_name: str) -> int:
+        query = DatasetQuery(query_type=QueryType.GET_ITEM_SIZE, item_name=item_name)
+        request = DatasetManagementMessage(action=ManagementAction.QUERY, query=query, dataset_name=dataset_name,
+                                           data_location=item_name)
+        self.last_response: DatasetManagementResponse = await self.async_make_request(request)
+        if self.last_response.success:
+            return self.last_response.data
+        else:
+            return -1
```
aaraney, line 223: ^^ ditto

python/lib/client/dmod/client/request_clients.py
```diff
@@ -175,6 +202,50 @@ async def delete_dataset(self, name: str, **kwargs) -> bool:
...
+        AnyStr
+            The downloaded block/chunk.
+        """
+        request = DatasetManagementMessage(action=ManagementAction.REQUEST_DATA, dataset_name=dataset_name,
+                                           data_location=item_name, blk_start=blk_start, blk_size=blk_size)
+        self.last_response: DatasetManagementResponse = await self.async_make_request(request)
+        if self.last_response.success:
+            return self.last_response.data
+        else:
+            return ''
```
aaraney, line 247: ^^ ditto. I am thinking for non-dict return types, does it make sense to raise an exception here?

robertbartel:
FWIW, I don't particularly understand why tests are (now) failing after rebasing in other upstream changes.  The same tests are not failing locally for me or when run on ucs3.  Given that, I think we should proceed with this (pending final review approval).

python/gui/MaaS/cbv/DatasetApiView.py
aaraney: ^^
Related, do we want to provide any context as to why a 400 was the response? i.e. expected one of `request_type`: <list of request types>.

python/gui/MaaS/cbv/DatasetManagementView.py
```diff
@@ -0,0 +1,114 @@
...
+        http_request : HttpRequest
+            The raw HTTP request in question.
+
+        Returns
+        -------
+        communication.MessageEventType
+            Either ``communication.MessageEventType.DATASET_MANAGEMENT`` or
+            ``communication.MessageEventType.DATA_TRANSMISSION``.
+        """
+        # TODO:
```
aaraney, line 42: Issue to track this? Or will this be completed in this PR?

python/gui/MaaS/cbv/DatasetManagementView.py
```diff
@@ -0,0 +1,114 @@
...
+from dmod.core.meta_data import DataCategory, DataFormat
+
+import logging
+logger = logging.getLogger("gui_log")
+
+from .utils import extract_log_data
+from .AbstractDatasetView import AbstractDatasetView
+
+
+class DatasetManagementView(AbstractDatasetView):
```
aaraney, line 18: Thoughts on adding a `delete` method here to handle deleting a dataset? Or will that fall under the responsibility of `post`? Meaning, `post` would then likely handle creation, deletion, and modification.

python/gui/MaaS/static/maas/js/components/confirmDialog.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+class ConfirmDeleteDatasetDialog extends ConfirmDialog {
+    constructor(dataset_name, parentDivName, id, styleClass, onConfirmFunc) {
+        super(parentDivName, id, styleClass, onConfirmFunc);
+        this.dataset_name = dataset_name;
+        this.buttons_div = null;
+    }
+    
+    _style_outer_div() {
```
aaraney, line 24: I think a lot of this styling will go away in the future, but for now it might make sense to leave it until we have the functionality from this PR and from #210. I would like to revisit the css that we are using GUI-wide in the nearish future.

python/gui/MaaS/static/maas/js/components/createDatasetForm.js
```diff
@@ -0,0 +1,42 @@
...
+
+        let addUploadSelection = false;
+        if (selection == "NETCDF_FORCING_CANONICAL") {
```
aaraney, line 16: As also noted in #198, in the future all format selections likely will have a file upload selection. Just leaving this here to relate this.

python/gui/MaaS/static/maas/js/components/createDatasetForm.js
```diff
@@ -0,0 +1,42 @@
+class CreateDatasetForm {
```
aaraney, line 1: I think we will be able to remove this after #210.

python/gui/MaaS/templates/maas/dataset_management.html
aaraney: This and we can share common things like stylesheets.

python/gui/MaaS/templates/maas/dataset_management.html
```diff
@@ -0,0 +1,797 @@
...
+    <link rel="stylesheet" type="text/css" href="https://ajax.googleapis.com/ajax/libs/jqueryui/1.12.1/themes/cupertino/jquery-ui.css">
+    <link rel="icon" type="image/gif" href="{% static 'maas/img/favicon.gif' %}">
+    <script>
+        function activate_model(model) {
+            $(".parameters").hide();
+            $("#" + model + "_parameters").show();
+        }
+
+        function buildDatasetsOverviewTable() {
+            $.ajax({
```
aaraney, line 26: In the future, we should avoid bringing in `jquery`. With the advent of the [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) API, it has become the preferred way to make requests.

python/gui/MaaS/templates/maas/dataset_management.html
```diff
@@ -0,0 +1,797 @@
...
+        }
+
+        function initGuiViews() {
+            buildDatasetsOverviewTable();
+            initDatasetCreateForm();
+        }
+
+        function changeCreateDatasetFormatSelection(selection) {
+            let dy_div = document.getElementById("dataset-create-form-div-dynamic-inputs");
+            for (let i = 0; i < dy_div.children.length; i++) {
```
aaraney, line 56: I think `dy_div.children[i].remove()` changes the length of `dy_div.children`. Instead, I think we want the following:

```javascript
// remove all dynamic input child elements
while (dy_div.firstChild) {
  dy_div.removeChild(dy_div.firstChild);
}
```

python/gui/MaaS/templates/maas/dataset_management.html
```diff
@@ -0,0 +1,797 @@
...
+            let format_select = document.getElementById("dataset-create-form-input-format");
+            format_select.selectedIndex = -1;
+        }
+
+        function initGuiViews() {
+            buildDatasetsOverviewTable();
+            initDatasetCreateForm();
+        }
+
+        function changeCreateDatasetFormatSelection(selection) {
```
aaraney, line 54: I think we can remove this after #210.

python/gui/MaaS/templates/maas/dataset_management.html
aaraney: I think we will be able to remove a lot of this if / when we move to `css` `grid` / `flexbox`. I think we are all on the same page though.

python/gui/MaaS/templates/maas/dataset_management.html
```diff
@@ -0,0 +1,797 @@
...
+        }
+
+        function makeByteSizesReadable(bytes, decimal_places = 2) {
+            if (bytes <= 0) {
+                return "0 B";
+            }
+            let thresh = 1024;
+            let decimals = decimal_places < 0 ? 0 : decimal_places;
+            let units = ['B', 'KB', 'MB', 'GB', 'TB'];
+            let level = Math.floor(Math.log(bytes) / Math.log(thresh))
```
aaraney, line 237: `Math.log` is the natural log. I believe we want [`Math.log2`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/log2).

aaraney: nevermind, I see you are doing change of base.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -460,6 +531,55 @@ async def delete_dataset(self, name: str, **kwargs) -> bool:
         self.last_response = await self.async_make_request(request)
         return self.last_response is not None and self.last_response.success
 
+    async def download_item_block(self, dataset_name: str, item_name: str, blk_start: int, blk_size: int) -> AnyStr:
```
aaraney, line 534: Likewise in this method. It would helpful to return some kind of Result type indicating success or failure and it might also be nice to return the number of bytes read / downloaded since the block size is maximum size to download.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -505,6 +625,43 @@ async def download_from_dataset(self, dataset_name: str, item_name: str, dest: P
...
+        action = ManagementAction.QUERY
+        query = DatasetQuery(query_type=QueryType.GET_SERIALIZED_FORM)
+        try:
+            for d in datasets:
+                request = MaaSDatasetManagementMessage(action=action, query=query, dataset_name=d,
+                                                       session_secret=self.session_secret)
+                self.last_response: DatasetManagementResponse = await self.async_make_request(request)
+                if self.last_response.success:
+                    serialized[d] = self.last_response.data[DatasetManagementResponse._DATA_KEY_QUERY_RESULTS]
+                # TODO: what to do if any are not successful
```
aaraney, line 659: I think we can avoid this more or less, or at least have less of an opinion about it if we move this method to an async generator rather than an accumulator. We could just yield results as we get them.

python/gui/MaaS/cbv/AbstractDatasetView.py
robertbartel: Essentially yes. In particular, to reusably abstract a view that has a dataset client property. But it's not especially meaningful on it's own.

python/gui/MaaS/cbv/DMODProxy.py
robertbartel: For some reason I thought I put that in settings.py.

christophertubbs:
> FWIW, I don't particularly understand why tests are (now) failing after rebasing in other upstream changes. The same tests are not failing locally for me or when run on ucs3. Given that, I think we should proceed with this (pending final review approval).

It must be a new version of `pint` being used. For some reason, the author changed up the alias for `Quantity` to point to something that *isn't* generic. The class we want now is `pint.PlainQuantity`, which *is* generic.

```python
  File "/home/runner/work/DMOD/DMOD/dmod_venv/lib/python3.8/site-packages/dmod/evaluations/measurement_units.py", line 54, in UnitConverter
    def get_quantity(self, value: _T, value_type: str) -> pint.Quantity[_T]:
```

just needs to be changed to:

```python
  File "/home/runner/work/DMOD/DMOD/dmod_venv/lib/python3.8/site-packages/dmod/evaluations/measurement_units.py", line 54, in UnitConverter
    def get_quantity(self, value: _T, value_type: str) -> pint.PlainQuantity[_T]:
```

robertbartel:
Closing; consolidating these and other changes in #240.

PR #199, opened 2022-10-19
Move Dataset classes to dmod.core package

robertbartel:
Moving the `Dataset` class and closely related, co-dependency types from _dmod.modeldata_ to _dmod.core_ to facilitate use of those classes in other places, in particular certain parts of the client and GUI.


robertbartel:
With PRs #197 and #199, I believe once one of them is merged, the other will need modifications to resolve conflicts.

christophertubbs (approved):

PR #208, opened 2022-10-27
Updated ngen worker image for NetCDF 

robertbartel:
Updated image to include support for NetCDF datasets.


aaraney (dismissed):
I don't have enough background yet on the ngen dependencies to notice anything out of the ordinary, but other than that, this all looks pretty sane to me.

mattw-nws (approved):
Successfully ran a formulation with NetCDF forcing with this revision. 👍

PR #210, opened 2022-10-28
Dynamic Dataset creation form (WIP)

### LINKED ISSUE #198: Update Dataset Creation Form For Dynamic Fields
The [dataset management GUI view](https://github.com/robertbartel/DMOD/blob/e21/dataset_mgr_gui/main/python/gui/MaaS/cbv/DatasetManagementView.py) (still under review #197) has a component view and form for creating new datasets. At the moment, the dynamic field for an upload input is partially broke (although that may need to be dynamic). This needs to be corrected. 

It also does not yet have anything for defining the [DataDomain](https://github.com/robertbartel/DMOD/blob/b52cfd7f03f56b4100b2632620cdb33e49a26e39/python/lib/core/dmod/core/meta_data.py#L423) of the dataset. There should be some dynamic fields for setting these during creation. They should depend on the [DataFormat](https://github.com/robertbartel/DMOD/blob/b52cfd7f03f56b4100b2632620cdb33e49a26e39/python/lib/core/dmod/core/meta_data.py#L35) selection, as this will determine the [indices](https://github.com/robertbartel/DMOD/blob/b52cfd7f03f56b4100b2632620cdb33e49a26e39/python/lib/core/dmod/core/meta_data.py#L9) applicable within the domain.

aaraney:
closes #198 

Once this is done, it will need to be merged after #197.


aaraney:
@robertbartel, do either the catchment id's or data id's have an expected format (thinking regex)? I ask so that I could add the expected patterns to their respective form fields.

robertbartel:
@aaraney, the catchment ids probably do: `cat-[0-9]+`.  I'm not sure how much that is official specification versus convention, though.

The data ids probably do not.  I have basically been using the dataset name as the data id values, although that's for human easy of use and not necessarily going to continue once things are more heavily utilized.  

It occurs to me that (object store) dataset names actually can't contain underscore characters, because their correspondence to bucket names in MinIO and MinIO's bucket name rules.

aaraney:
> @aaraney, the catchment ids probably do: `cat-[0-9]+`. I'm not sure how much that is official specification versus convention, though.

Thanks for clarifying, I am going to take that to mean we should hold off restricting "valid" catchment ids, at least for now.

> The data ids probably do not. I have basically been using the dataset name as the data id values, although that's for human easy of use and not necessarily going to continue once things are more heavily utilized.

👍. It seems like in the future dataset id's might be something we will generate ourselves. Is that fair?

> It occurs to me that (object store) dataset names actually can't contain underscore characters, because their correspondence to bucket names in MinIO and MinIO's bucket name rules.

Cool cool. I will add filename restriction to avoid that.

aaraney:
I added filename and start / end time validation, meaning you cannot submit the form without these fields set properly. Invalid form entry shows meaningful error messages so the person can make corrections.

robertbartel:
Closing; consolidating these and other changes in #240.

PR #220, opened 2022-11-22
Adding Grouped Single-Node Allocation Paradigm

robertbartel:
Adding support for `AllocationParadigm.GROUPED_SINGLE_NODE`.  This represents associating all requested CPU and memory with a single _ResourceAllocation_ object, which eventually translates to a single Docker container with all the CPU/memory.  

This new paradigm is in contrast to the existing values - in particular  _SINGLE_NODE_ - which allocate 1 CPU and identical fractions of memory to `n` _ResourceAllocation_ objects (i.e., 1 CPU and similar portions of memory to each container), where `n` is the total number of requested CPUs.

### DIFF
--- a/python/lib/core/dmod/core/execution.py
+++ b/python/lib/core/dmod/core/execution.py
@@ -15,11 +15,13 @@ class AllocationParadigm(Enum):
                       allocations be single cpu/process
         SINGLE_NODE - require all allocation of assets to be from a single resource/host; also, require allocations to
                       be single cpu/process
+        GROUPED_SINGLE_NODE     -   obtain one grouped allocation of required assets, all from a single resource/host
     """
 
     FILL_NODES = 0
     ROUND_ROBIN = 1
     SINGLE_NODE = 2
+    GROUPED_SINGLE_NODE = 3
 
     @classmethod
     def get_default_selection(cls) -> 'AllocationParadigm':
--- a/python/lib/scheduler/dmod/scheduler/job/job_manager.py
+++ b/python/lib/scheduler/dmod/scheduler/job/job_manager.py
@@ -632,6 +632,8 @@ class RedisBackedJobManager(JobManager, RedisBackedJobUtil):
             alloc = self._resource_manager.allocate_fill_nodes(job.cpu_count, job.memory_size)
         elif job.allocation_paradigm == AllocationParadigm.ROUND_ROBIN:
             alloc = self._resource_manager.allocate_round_robin(job.cpu_count, job.memory_size)
+        elif job.allocation_paradigm == AllocationParadigm.GROUPED_SINGLE_NODE:
+            alloc = self._resource_manager.allocate_grouped_single_node(job.cpu_count, job.memory_size)
         else:
             alloc = [None]
         if isinstance(alloc, list) and len(alloc) > 0 and isinstance(alloc[0], ResourceAllocation):
--- a/python/lib/scheduler/dmod/scheduler/resources/resource_manager.py
+++ b/python/lib/scheduler/dmod/scheduler/resources/resource_manager.py
@@ -231,6 +231,30 @@ class ResourceManager(ABC):
         # Otherwise, return the allocations
         return allocations
 
+    def allocate_grouped_single_node(self, cpus: int, memory: int) -> List[Optional[ResourceAllocation]]:
+        """
+        Check available resources to allocate job request to a single allocation (on a single node) with all resources.
+
+        Parameters
+        ----------
+            cpus: Total number of CPUs requested
+            memory: Amount of memory required in bytes
+
+        Returns
+        -------
+        List[Optional[ResourceAllocation]]
+            List containing a single element, where this element is either a ::class:`ResourceAllocation` if allocation
+            was successful, or ``None``.
+        """
+        self.validate_allocation_parameters(cpus, memory)
+
+        for r in self.get_useable_resources():
+            if r.cpu_count >= cpus and r.memory >= memory:
+                alloc = self.allocate_resource(resource_id=r.resource_id, requested_cpus=cpus, requested_memory=memory)
+                if alloc is not None:
+                    return [alloc]
+        return [None]
+
     def allocate_round_robin(self, cpus: int, memory: int) -> List[Optional[ResourceAllocation]]:
         """
             Check available resources on host nodes and allocate in round robin manner even the request
--- a/python/lib/scheduler/dmod/scheduler/scheduler.py
+++ b/python/lib/scheduler/dmod/scheduler/scheduler.py
@@ -659,6 +659,7 @@ class Launcher(SimpleDockerUtil):
                 service_params.user = 'root'
                 # Also adding this for ngen
                 service_params.capabilities_to_add = ['SYS_ADMIN']
+            # TODO: actually impose/enforce resource allocations on created services
             #TODO check for proper service creation, return False if doesn't work
             service = self.create_service(serviceParams=service_params, idx=alloc_index,
                                           docker_cmd_args=self._generate_docker_cmd_args(job, alloc_index))


robertbartel:
@aaraney, that's the immediate motivation, but this will probably have more general use. There will also likely be "grouped" analogs for the two other existing values, again, for added flexibility (and perhaps greater efficiency) of execution. 

I've no doubt we will want to expand and perhaps overhaul the design in the future. We will need to be careful how much control/responsibility we give to users though, and that we don't overwhelm those who are less interested in how the computation is distributed.

hellkite500:
I think there is some confusion on what `SINGLE_NODE` does and/or is expected to do.  I'm pretty sure it will allocate multiple CPU's, but they must all be on the same NODE:

```python
    def allocate_single_node(self, cpus: int, memory: int) -> List[Optional[ResourceAllocation]]:
        """
        Check available resources to allocate job request to single-cpu allocations on a single node.
        Parameters
        ----------
            cpus: Total number of CPUs requested
            memory: Amount of memory required in bytes
        Returns
        -------
        [ResourceAlloction]
            List of ResourceAllocation if allocation successful; otherwise, [None]
        """
        #Fit the entire allocation on a single resource
        self.validate_allocation_parameters(cpus, memory)
        for res in self.get_useable_resources():
            if res.cpu_count >= cpus and res.memory >= memory:
                allocations = []
                mem = int(memory / cpus)
                for i in range(cpus):
                    alloc = self.allocate_resource(resource_id=res.resource_id, requested_cpus=1, requested_memory=mem)
                    if alloc is None:
                        allocations = None
                        break
                    else:
                        allocations.append(alloc)
                if allocations is not None:
                    return allocations
        return [None]
```

Though this is at odds with a few different doc strings.  The original implementation and intent of SINGLE_NODE was just that -- provide multiple CPUS's on a single node (thus no need for distributed communication) to support exactly the kind of threaded/multiprocessing behavior the calibration use case presents.

robertbartel:
@hellkite500, you are correct on the original intent, but that was changed. Reintroducing something to represent that is why this is needed. 

The change was largely, if not totally, to adhere to the Docker principle of one process per container (and thus the need for only one CPU).  This was done for all the existing paradigms. 

How much we permanently support things that don't adhere to this principle will probably come down to external constraints and performance comparisons, but for now we want both (at least for the single-host case).

robertbartel:
To be clear, this will indeed allocate multiple CPUs, and they will be all on the same physical host. But it will do so by creating a collection of `ResourceAllocation` objects, each requested to have `1` CPU. These will then directly correspond to the _ngen_ worker containers started by the `Launcher` object within the scheduler service.

hellkite500:
> To be clear, this will indeed allocate multiple CPUs, and they will be all on the same physical host. But it will do so by creating a collection of `ResourceAllocation` objects, each requested to have `1` CPU. These will then directly correspond to the _ngen_ worker containers started by the `Launcher` object within the scheduler service.

This is what I wasn't clear on.  The idea being multiple service containers can be used to serve a single `job`, and we ensure each of those containers is allocated `1` virtual CPU from our virtual resource pool, as opposed to a single job using a single container but using multiple virutal CPU's.

aaraney:
Sorry if I am kind of making a separate thread. I wasn't the most concise in my initial comment. What I didn't say, but was trying to convey, was that it seems like `AllocationParadigm` has blended imperative and declarative forms of allocation. For example,  the round robin scheduler specifies _how_ to schedule work, but purposefully doesnt constrain _where_. In contrast, the grouped single node scheduler is constraining both _how_ and _where_ to schedule work. I think this might lead to a little confusion, like @hellkite500 has mentioned, and _potentially_ lead to single and multinode versions of some paradigms like @robertbartel alluded to with this:

> There will also likely be "grouped" analogs for the two other existing values, again, for added flexibility (and perhaps greater efficiency) of execution.

I \*think\* we can gain some nice characteristics if we are to further separate the _declarative_ aspects of scheduling (what resources are requested; i.e. 1 node, 4 replicas, 1 vCPU per/replica) from the _imperative_ aspects (_which_ resources are allocated). By doing so, I think the majority of existing `AllocationParadigm`s can be handled by a single `AllocationParadigm`. I still need to work through this more, but am I at least making more sense now?

robertbartel:
> For example, the round robin scheduler specifies how to schedule work, but purposefully doesn't constrain where. In contrast, the grouped single node scheduler is constraining both how and where to schedule work.

I don't agree.  

Round robin does, in general, constrain _where_ work is distributed:  _**everywhere**_.   It is the only one that does this, though.   For the others, what matters is the order in which nodes are examined - which does not depend on the paradigm (I think this is effectively random) - and how many compute assets (i.e., CPU and memory ***) remain on the "current" node.

All paradigms address _how_ work is distributed.  Round robin is **_evenly_**, single node is _**all together or not at all**_, and fill nodes is _**as much together as possible**_.

So, I'm not sure if I follow exactly what you have in mind for the design.  But I do agree that there is plenty of room for enhancing this in the future.

\*\*\* We may want/need to have other types of compute assets in the future beyond CPU and memory, such as storage space or GPUs.  And when we do, that surely will have implications to how we need to address allocations.

aaraney:
> @aaraney, @hellkite500, we seem to have lost track of this one as other things came up at the end of 2022. If my understanding of the functionality of **ngen-cal** is correct, we will need this change for the calibration job workflow. That workflow, in turn, will need more active development and testing in the near term.
> 
> Were there still outstanding questions or discussions on this that needed to be resolved?

Thanks for revitalizing this, @robertbartel. And for your discussion above. 

Revisiting my prior comment:

> Sorry if I am kind of making a separate thread. I wasn't the most concise in my initial comment. What I didn't say, but was trying to convey, was that it seems like `AllocationParadigm` has blended imperative and declarative forms of allocation. For example, the round robin scheduler specifies _how_ to schedule work, but purposefully doesnt constrain _where_. In contrast, the grouped single node scheduler is constraining both _how_ and _where_ to schedule work. I think this might lead to a little confusion, like @hellkite500 has mentioned, and _potentially_ lead to single and multinode versions of some paradigms like @robertbartel alluded to with this:

and @robertbartel's response, what I failed to say more plainly was, are the intent of _some_ `AllocationParadigm` variants more clearly expressed by introducing compute assets `n_tasks` and / or `n_nodes`? I am not advocating for adding `n_tasks` and `n_nodes` to all `AllocationParadigm` variants, just variants' where the paradigm is more clearly captured with the inclusion of compute asset bounds, `n_tasks` and / or `n_nodes`. A good use case for this functionality is testing, debugging, and constrained deployment environments. I assume that exclusion of `n_tasks` and `n_nodes` thus far has been to purposefully abstract these concepts from the user and by doing so simplify the api. There may be cases in the future where we want the ability to further constrain the scheduler's scheduling bounds.

robertbartel:
> ... are the intent of some _AllocationParadigm_ variants more clearly expressed by introducing compute assets `n_tasks` and / or `n_nodes`?

No, I don't think so.  The paradigm selection should just be the general pattern for the scheduler to use to build the resource allocations, not detailed constraints. 

> I am not advocating for adding n_tasks and n_nodes to all AllocationParadigm variants, just variants' where the paradigm is more clearly captured with the inclusion of compute asset bounds, n_tasks and / or n_nodes

That lack of ubiquity is directly related to why those details belong somewhere else. 

> A good use case for this functionality is testing, debugging, and constrained deployment environments

I am interested in you elaborating on this, but I expect test cases would be more appropriately handled by just manually initializing resources pools in a test _ResourceManager_ and/or _ResourceAllocation_ objects.

> I assume that exclusion of n_tasks and n_nodes thus far has been to purposefully abstract these concepts from the user and by doing so simplify the api.

Essentially, but it's important to be clear that, between _making things easier for the user_, and _users not having fine-grained scheduling control_, the latter is actually the higher priority here.

> There may be cases in the future where we want the ability to further constrain the scheduler's scheduling bounds.

Again, DMOD itself can, but users should not be able to constrain the scheduler directly except for very limited exceptions that don't offer precise control.

It is useful and/or necessary to give some control levers to users, which we have:  options to communicate either a strict or loose requirement that all job processing be on one physical host, or to indicate all job processing resources are required to be within a single container instead of several.  Those are directly motivated by use cases of the software that gets executed in DMOD jobs.

robertbartel:
Just to add to this a little further, I feel like I could be missing part of what @aaraney has been trying to describe here, and I want to make sure everything is clear before we consider this conversation finished.  However, I think we've gotten out of scope for this PR.  

If I am also missing a reason why something being discussed should give us pause on adding these changes, let me know.  Otherwise, I suggest we separate the more detailed design discussion to a separate issue.

aaraney:
@robertbartel, thanks for guiding the conversation. I agree, I think my discussion points are a little out of scope and don't directly affect this PR. I may however open an issue to continue discussion of this.

aaraney:
Revisiting this, I think the point I failed to adequately convey is out of scope for this PR.

However, in general I think it would be beneficial to, in the future, add an intermediary `GenericResourceAllocation` or something similar. It appears that all of the current allocation paradigms could be generally described by the number of nodes, tasks, and amount of memory requested _if_ the compute topology is known. 

If this is true, it seems a generic scheduler could be written, such that, given a `GenericResourceAllocation` it returns one or more `ResourceAllocation`s. So, an allocation paradigm would just return a `GenericResourceAllocation` that the resource allocation manager would turn into one or more `ResourceAllocation`s. By doing this, we would centralize the resource allocation code into a single unit (this has nice traits if we needed to support non-docker resources in the future) and it would let us:
- fail faster because we would know if we could fulfill a request before trying to fulfill it, - enable more informative error messages that dont necessarily need to be obfuscated, 
- and assist in debugging (i.e. reproducing a request on a system with different topology).

PR #222, opened 2022-11-23
Add factory method for creating DataDomain

robertbartel:
Adding classmethod to factory create a `DataDomain` from a `DataFormat` and a collection of keyword args representing the constraints of the domain.

This will be used by some GUI work currently being developed, in particular for creating datasets.  However, it stands alone cleanly (and will probably find more general use), so separating it for an easier review process.

### DIFF
--- a/python/lib/core/dmod/core/meta_data.py
+++ b/python/lib/core/dmod/core/meta_data.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from .serializable import Serializable
 from numbers import Number
 from typing import Any, Dict, List, Optional, Set, Type, Union
+from collections.abc import Iterable
 
 
 class StandardDatasetIndex(Enum):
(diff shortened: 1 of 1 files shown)


python/lib/core/dmod/core/meta_data.py
```diff
@@ -467,6 +468,75 @@ def factory_init_from_deserialized_json(cls, json_obj: dict):
...
 
+    @classmethod
+    def factory_init_from_restriction_collections(cls, data_format: DataFormat, **kwargs) -> 'DataDomain':
+        """
+        Create and return a data domain object of the given format and keyword args containing restriction properties.
+
+        The key for each restriction determines the appropriate ::class:`StandardDatasetIndex` for the restriction's
+        ``variable`` property via ::method:`StandardDatasetIndex.get_for_name``.
+
+        The restriction property values should either be a single value, a list or a dictionary with exactly two inner
```
aaraney, line 479: This is super minor, but I think using an oxford comma here makes this a little more clear:

> [...] single value, a list, or a dictionary [...].

aaraney (approved):
I left one minor grammar suggestion. Otherwise, this looks good to me! Thanks, @robertbartel!

python/lib/core/dmod/core/meta_data.py
aaraney: Nice! Thanks for doing that. Re-reviewing / approving now.

aaraney (approved):

PR #226, opened 2022-12-28
Fix build for new pytables dependency on blosc2

mattw-nws:
Fix build for new pytables dependency on blosc2

## Additions

- Explicit installs of Cython and blosc2 in the deps image (Cython should have been installed as a dependency elsewhere before).

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [X] PR has an informative and human-readable title
- [X] Changes are limited to a single goal (no scope creep)
- [X] Code can be automatically merged (no conflicts)
- [X] Code follows project standards (link if applicable)
- [X] Passes all existing automated tests
- [X] Any _change_ in functionality is tested
- [X] New functions are documented (with a description, list of inputs, and expected output)
- [X] Placeholder code is flagged / future todos are captured in comments
- [X] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [X] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [X] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Target Environment support

- [X] Windows
- [X] Linux
- [X] Browser

### Accessibility

- [X] Keyboard friendly
- [X] Screen reader friendly

### Other

- [ ] No linting errors or warnings

### DIFF
--- a/docker/main/ngen/Dockerfile
+++ b/docker/main/ngen/Dockerfile
@@ -346,7 +346,7 @@ RUN echo "export HYDRA_HOST_FILE=${HYDRA_HOST_FILE}" >> /etc/profile \
     # && ctest \
     && make install \
     # Install required python dependency packages with Pip \
-    && pip3 install numpy pandas pyyaml bmipy netCDF4 wheel packaging \
+    && pip3 install numpy pandas pyyaml bmipy Cython blosc2 netCDF4 wheel packaging \
     && HDF5_DIR=/usr pip3 install --install-option="\'--hdf5 \${HDF5_DIR}\'" --install-option="\'--jobs=$(nproc) \'"  --no-build-isolation tables \
     # Make aliases for convenience \
     && alias pip='pip3' \


mattw-nws:
Good questions. This fix was somewhat reactionary so there could be a better place for some of this. Here's the rub though...

These are here now so that pytables will build--from source--properly. So the question may be whether the building of pytables itself should be moved. Problem is that this pytables build requires the HDF5 libraries from ngen-deps, so that the ones in pytables will match the ones in ngen. And, this built pytables has to get into the python environment somehow...

As for that latter part, it might be possible to build a wheel of pytables with the necessary dependencies over in the `rocky_build_troute` intermediate stage, and then put that wheel in with the four t-route ones and install it... but I'm not sure that works. At least `rocky_build_troute` does derive from `rocky-ngen-deps`, so the HDF5 libs should be there. Should I try this?

robertbartel:
> These are here now so that pytables will build--from source--properly. So the question may be whether the building of pytables itself should be moved. Problem is that this pytables build requires the HDF5 libraries from ngen-deps, so that the ones in pytables will match the ones in ngen. 

Earlier I was incorrectly assuming (after searching for "pytables") that the **pytables** build and install was hidden somewhere in the **t-route** steps, but I see now.  I think the HDF5 version would be fine though, as long as a stage deriving from `rocky-ngen-deps` is used, as you suggest below.

That said, now I find myself asking what actually requires **pytables**.  Once again, I assumed **t-route**, but **tables** is not in its _requirements.txt_.  

> And, this built pytables has to get into the python environment somehow...

I presume this is the crux of the problem.  Unless we create another stage, either **pytables** (and these new explicit dependencies) go into something more general, or they are missing from the final stage of the image (which copies from `rocky_build_troute`, but does not derive from it).

> As for that latter part, it might be possible to build a wheel of pytables with the necessary dependencies over in the `rocky_build_troute` intermediate stage, and then put that wheel in with the four t-route ones and install it... but I'm not sure that works. At least `rocky_build_troute` does derive from `rocky-ngen-deps`, so the HDF5 libs should be there. Should I try this?

The `py-sources` stack is essentially for this:  pre-build necessary packages, including both internal DMOD packages and dependencies that need to be build from source.  To answer your question, no, not as part of this PR.  For now, your original change will suffice, so I will approve it shortly.  But I will open an issue to do that work later.

mattw-nws:
> The `py-sources` stack is essentially for this: pre-build necessary packages, including both internal DMOD packages and dependencies that need to be build from source

Ah--this is where this gets hairy. This special pytables build is unique to ngen. DMOD itself could use a different version and/or doesn't care what HDF5 library it's built with. The problem that this particular pytables build was designed to solve is one introduced by the use of pybind11 in ngen--the HDF5 library apparently has to match the one built into ngen or crashes happen. For the (mostly-pure?) Python stack in DMOD, it can use a different pytables with a different HDF5 library (by default, installing pytables as a binary wheel brings along a recent HDF5 .so file that the wheel is built against--which was the problem!). So, I think py-sources is the wrong place to put this... but I could be wrong and/or misunderstanding py-sources.

mattw-nws:
(addendum to above)
It would also require the HDF5 libraries built in `rocky-ngen-deps` or a rebuild of the exact same ones... and I don't think we'd want to reference the ngen deps image in the DMOD portion or mix the two things up like that? Or would we?

robertbartel:
> (addendum to above) It would also require the HDF5 libraries built in `rocky-ngen-deps` or a rebuild of the exact same ones... and I don't think we'd want to reference the ngen deps image in the DMOD portion or mix the two things up like that? Or would we?

It would require identical libraries in both places, yes.  That would probably mean an image in the `py-sources` stack based on the Rocky Linux image, but migrating to that doesn't sound like a terrible thing to do anyway.   And we wouldn't want `py-sources` to depend on `rocky-ngen-deps` in any way, but there are other ways to ensure both end up with the same HDF5 library.

You raise a good point about not mixing things, but I think we could still make something like that work.

robertbartel (approved):

PR #239, opened 2023-01-10
Refactor dmod.core serialization and deserialization types

aaraney:
This PR introduces major changes to the way `dmod.core` handles serialization and deserialization. Namely, serialization, deserialization, validation, and type coercion are now _mainly_ handled by [`pydantic`](https://docs.pydantic.dev/usage/models/#basic-model-usage).
All subtypes of `dmod.core.serializable.Serializable` are affected by this change. This PR specifically introduces the changes to `dmod.core.serializable.Serializable` and refactored implementations of all `Serializable` subtypes in the `dmod.core` namespace. This PR also introduces a new `enum.Enum` variant, `dmod.core.enum.PydanticEnum`, that exposes enum's fields consistently when embedded in a `Serializable` subtype. A following PR will provide refactored implementations for `dmod.communication`.

Given the large scope of this change, I've broken this PR description up into sections:

1. [Motivation](#Motivation)
2. [Implications](#Implications)
3. [Examples and Diagrams](#Examples-and-Diagrams)

The examples and diagrams section is specifically included to assist developers in getting up to speed and becoming acquainted with the new implementation of `dmod.core.serializable.Serializable`. The intent of the examples and diagrams section is also to be used a as a reference.

There are several locations in the code where I still have questions that will need to be resolved before merging this PR. I've done my best to comment and tag you all where these questions have come up. 

<h3 id="Motivation">Motivation</h3>

The _primary goal_ of this PR is to enhance the `Serializable` type by removing room for inconsistency and improving code comprehension, readability, and maintainability.

Writing serialization and deserialization code is a laborious and repetitive task. Writing easily comprehensible, consistent, readable, and maintainable serialization and deserialization code often requires a higher degree of diligence and consistency than the application's business logic from both the roles of the programmer and reviewer(s). The code itself is not be hard to write, but ensuring that assumptions, readability, type coercion strategies, documentation, and exceptions are all consistent across a large code base requires careful eyes and attention to detail.

`dmod`'s existing serialization and deserialization implementations rely on a common parent class, `Serializable`. The `Serializable` class provides clear abstractions for accomplishing serialization and deserialization through the `to_dict` / `to_json` and `factory_init_from_deserialized_json` methods respectively. These api's are clear and effective, however a subtype's accepted or provided deserialized json structure and encoded assumptions are not always clear. Likewise, subclasses must implement a number of methods and properties (e.g. `__init__`, `__eq__`, `factory_init_from_deserialized_json`, `to_dict`) that leave room for inconsistency. Enhancements made in this PR to the `Serializable` type specifically address these pain points.

<h3 id="Implications">Implications</h3>

With the enhancements to the `Serializable` type, subtypes will no longer _need_ to implement 
`factory_init_from_deserialized_json`, `to_dict`, `__init__`, `__eq__`, or any field properties. However, the most glaring difference is how subtypes define their fields. Public fields are now defined as class variables along with their associated type hint.
As an added bonus, all subtypes now have a `schema_json` method that generates the associated json schema. This mean, in the future we will be able to leverage this feature to generate open-api documentation.

**User Type Example**
```python
from dmod.core.serializable import Serializable
from pydantic import EmailStr
from typing import Optional

class User(Serializable):
  id: int
  name: str
  email: Optional[EmailStr]

>>> user = User.factory_init_from_deserialized_json({"id": 12, "name": "test user"})
>>> user.to_dict() # {"id": 12, "name": "test user"} # notice optional field, email, is excluded because it is None
```

The `Serializable` type will try and coerce input into the specified field's type hint type. Meaning, `User.factory_init_from_deserialized_json({"id": "12", "name": "test user"})` is okay and the `id` field will be correctly coerced into an `int`.

Other notable implications:

- Fields with value `None` (i.e. `Optional[T]`) are excluded from both `to_dict` and `to_json`. This is consistent with prior `to_dict` implementations. This does not include 'unset' fields or non-`None` fields (i.e. empty list).
- Fields are serialized using their [alias](https://docs.pydantic.dev/usage/model_config/#alias-precedence) name if provided.
- `Serializable` subtypes fields can be instantiated using either their field name or alias name.
- Subtypes are serialized by calling [`dict`](https://docs.pydantic.dev/usage/exporting_models/#modeldict) -> `to_dict` -> `to_json`. If a subtype needs special serialization, overload the `dict` method, **not** `to_dict`.
- Class variables are now defined by providing a field the `typing.ClassVar[T]` type hint.
- Private fields are defined either by setting the field `=` to `pydantic.PrivateVar()`, or by adding a single leading `_`. These fields are not included in initializer (`__init__`).

#### Enums

A new `enum.Enum` variant, `dmod.core.enum.PydanticEnum`, has been added. This exposes enums correctly and consistently when they are embedded as fields in a `Serializable` subtype. Enums are serialized using their `name` property and validated using their `name` property _case insensitively_. Likewise, enums are exposed in `Serializable` json schema's using their `name` property in upper case.

```python
from dmod.core.enum import PydanticEnum
from dmod.core.serializable import Serializable

class PowerState(PydanticEnum):
  OFF = 0
  ON = 1

class Appliance(Serializable):
  power_state: PowerState

>>> Appliance(power_state=PowerState.OFF)
>>> Appliance(power_state="OFF")
>>> # enums are validated using their `name` property case insensitively
>>> Appliance(power_state="off")

>>> Appliance(power_state=0) # invalid

>>> # enums are serialized using their `name` property
>>> Appliance(power_state="off").to_dict() # {"power_state": "OFF"}
```

<h3 id="Examples and Diagrams">Examples and Diagrams</h3>

This section aims to get developer up to speed with using the enhanced `Serializable` subtype. As such, ive laid it out in the following subsections:

- [Inheritance and Composition](#Inheritance-and-Composition)
- [Three things to know](#Three-things-to-know])
- [Serialization and Deserialization](#Serialization-and-Deserialization)
- [Other helpful things](#Other-helpful-things])

<details>
<summary><h4>Inheritance and Composition</h4></summary>

To start, it is helpful to know how inheritance vs. composition behaves. By inheriting from another `Serializable` subtype, it is like adding its fields next to the ones you specify without introducing hierarchy. In contrast, composing or embedding another `Serializable` subtype will create hierarchy. For example:

```python
from dmod.core.serializable import Serializable

class Foo(Serializable):
  foo: str

class Bar(Foo):
  bar: str

class Baz(Serializable):
  quox: Foo


>>> Bar.factory_init_from_deserialized_json({"foo": "aaa", "bar": "bbb"})

>>> Baz.factory_init_from_deserialized_json({"quox": {"foo": "aaa"}}) # note the hierarchy
```

In the next section, I'll discuss validators, but for now just remember they are called on the subtype they are defined on _even if_ the subtype is inherited from or composed in another subtype. 

</details>

<details>
<summary><h4>Three things to know</h4></summary>

To be productive you will need to know about the following from `pydantic`:

- [`Field`](https://docs.pydantic.dev/usage/schema/#field-customization) - Optionally, provide extra information about the field and validations.
- [`validator`](https://docs.pydantic.dev/usage/validators/) - define custom field validation.
- [`root_validator`](https://docs.pydantic.dev/usage/validators/#root-validators) - define custom validation with access to all fields.

The main uses of `Field` are to add an `alias`, `description`, `default` or callable `default_factory`, or further constrain a type during validation.

```python
from dmod.core.serializable import Serializable
from typing import Optional
from random import randint

class Foo(Serializable):
  # field, foo, is serialized as `FOO`
  foo: str = Field(alias="FOO")

  # field has default, 42. constrain input between 0 and 50.
  bar: int = Field(42, ge=0, le=50)

  # field is not required, if not provided, get default from `default_factory`
  baz: bool = Field(default_factory=lambda _: bool(randint(0, 1))) 
```

Validators are extremely powerful and enable adding custom validation before or after pydantic does its validation. You can also define two validators and validate before pydantic and after, for example. Take the example of ensuring a file exists:

```python
from dmod.core.serializable import Serializable
from pydantic import validator
from pathlib import Path


class ModelConfig(Serializable):
  config_file: Path
  another_config_file: Path

  @validator("config_file", "another_config_file")
  def _validate_config_file_exists(cls, value: Path) -> Path:
    if not value.exists():
      # pydantic will catch this and propagate a pydantic.ValidationError
      raise FileNotFoundError(f"could not find file: {value}")
    
    return value

```

One thing to note is, by default, validators will not run if the field is not provided during construction (i.e. the field has a default). You can get around this by using the `always=True` flag in [`validator`](https://docs.pydantic.dev/usage/validators/). This can be a little awkward if you are unaware of this behavior. Take the following:

```python
from dmod.core.serializable import Serializable
from pydantic import validator
from typing import Optional

class Foo(Serializable)"
  bar: Optional[int] = 42

  @validator("bar")
  def _validate_bar(cls, value: Optional[int]) -> Optional[int]:
    return value

>>> Foo() # _validate_bar is not called
>>> Foo(bar=12) # " "     is called

>>> # Note this case
>>> Foo(bar=None) # " "   is called

```

What this means is, you should only use `Optional[T]` when you are okay with users providing `None` as a valid input. If the field is optional and has a sane default, mark it as such (i.e. `bar: int = 42`). I encourage you to read the [pydantic validator docs](https://docs.pydantic.dev/usage/validators/) to become familiar with the accepted arguments.

`root_validator` is very similar to validator, however instead of just validating a single field, the entire model (field and values) are passed as a dictionary to the `root_validator`. Take the example of verifying that two lists fields have at least one value overall: 

```python
from dmod.core.serializable import Serializable
from pydantic import root_validator
from typing import Dict, List

class HasAtLeastAValue(Serializable):
  l1: List[int]
  l2: List[int]

  @root_validator()
  def _validate_has_at_least_a_value(cls, values: Dict[str, List[int]]) -> Dict[str, List[int]]:
    assert len(values["l1"]) + len(values["l2"]) > 0
    return values

```

Like `validator`, `root_validator` comes in handy when fields are dependent on one other or computed field from another field. I would encourage you to become familiar with the [pydantic `root_validator` docs](https://docs.pydantic.dev/usage/validators/#root-validators) as well. 

</details>

<details>
<summary><h4>Serialization and Deserialization</h4></summary>

Below are code snippets for how serialization and deserialization are implemented. The main takeaway is, if a subtype needs some kind of custom serialization, overload `dict`.

##### Serialization

The thing to notice is, when calling `to_dict` or `to_json` you are first calling `dict`. If a subtype needs some kind of custom serialization, overload `dict`.

```python
class Serializable(BaseModel, ABC):
    def to_dict(self) -> Dict[str, Union[str, Number, dict, list]]:
        # see: https://docs.pydantic.dev/usage/exporting_models/#modeldict
        return self.dict(exclude_none=True, by_alias=True)

    def to_json(self) -> str:
        return json.dumps(self.to_dict(), sort_keys=True)
```

##### Deserialization

```python
class Serializable(BaseModel, ABC):
    @classmethod
    def factory_init_from_deserialized_json(cls: Self, json_obj: dict) -> Optional[Self]:
        try:
            # this is equivalent to Serializable(**json_obj).
            # this is helpful to know.
            return cls(**json_obj)
        except:
            return None
```

</details>

<details>
<summary><h4>Other helpful things</h4></summary>

It may also be helpful to know about:

- [Model Configuration](https://docs.pydantic.dev/usage/model_config/)
- [pydantic field types](https://docs.pydantic.dev/usage/types/#pydantic-types) (i.e. `pydantic.UUID4`)
- [strict types](https://docs.pydantic.dev/usage/types/#strict-types)
- [sequence of a type - custom root types](https://docs.pydantic.dev/usage/models/#custom-root-types)

</details>

## Additions

- New `enum.Enum` variant, `dmod.core.enum.PydanticEnum`, added that exposes enums fields consistently when embedded in a `Serializable` subtype. If you are implementing an enum and think it will be embedded in a `Serializable` subtype, subtype `PydanticEnum` instead of `Enum`.

## Changes

- Serialization, deserialization, validation, and type coercion are now _mainly_ handled by [`pydantic`](https://docs.pydantic.dev/usage/models/#basic-model-usage).
All subtypes of `dmod.core.serializable.Serializable` are affected by this change.
- With the enhancements to the `Serializable` type, subtypes will no longer _need_ to implement 
`factory_init_from_deserialized_json`, `to_dict`, `__init__`, `__eq__`, or any field properties. If you need special serialization, overload the new `dict` method on a `Serializable` subtype.
- `Serializable` subtype fields are now defined as class variables along their associated type hint (i.e. `field: int`). Use `typing.ClassVar[T]` to define a class variable or `pydantic.PrivateVar()` to define a private variable.
- `Serializable` subtypes now have a `schema_json` method that generates their associated json schema.  

## Testing

1. Verified that all `dmod.core` unit tests, except `test_allocationparadigm.py` pass locally. This and other tests fail because of the changes to `Serializable` that propagate to subtypes. If you would like me to resolve this issue for this PR, please let me know. The following `dmod.communication` PR will resolve downstream issues.

## TODO

- [ ] bump `dmod.core` patch version before merge.

## Checklist

- [x] Passes all existing automated tests
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)


python/lib/core/dmod/core/dataset.py
```diff
@@ -59,85 +59,93 @@ class Dataset(Serializable):
...
-                       is_read_only=json_obj[cls._KEY_IS_READ_ONLY],
-                       expires=cls._date_parse_helper(json_obj, cls._KEY_EXPIRES),
-                       derived_from=json_obj[cls._KEY_DERIVED_FROM] if cls._KEY_DERIVED_FROM in json_obj else None,
-                       derivations=json_obj[cls._KEY_DERIVATIONS] if cls._KEY_DERIVATIONS in json_obj else [],
-                       created_on=cls._date_parse_helper(json_obj, cls._KEY_CREATED_ON),
-                       last_updated=cls._date_parse_helper(json_obj, cls._KEY_LAST_UPDATE))
-        except Exception as e:
+    _SERIAL_DATETIME_STR_FORMAT: ClassVar = '%Y-%m-%d %H:%M:%S'
+    name: str = Field(description="The name for this dataset, which also should be a unique identifier.")
+    # QUESTION: should this be optional? see factory_init_from_deserialized_json
```
aaraney, line 64: @robertbartel, can you please weigh in on this question. See [this](https://github.com/NOAA-OWP/DMOD/pull/239/files#diff-bf7ff51a1703a364c9468cee4a087e6334748b16b1b6347a4f16332079b30786R123-R143) section of code for reference.

python/lib/core/dmod/core/dataset.py
```diff
@@ -335,8 +228,10 @@ def extend_life(self, value: Union[datetime, timedelta]) -> bool:
         if not self.is_temporary:
             return False
         elif isinstance(value, timedelta):
-            self._set_expires(self._expires + value)
+            # TODO: Fix bug. expires could be None
```
aaraney, line 231: @robertbartel should `self._expire` have a default value?

python/lib/core/dmod/core/dataset.py
robertbartel: Not sure I understand. Are you asking if the name should be an optional property?

python/lib/core/dmod/core/dataset.py
robertbartel: It should have a default value of `None`, if the dataset is not temporary. I'm not entirely sure whether that's what you were asking, or if you meant a default _datetime_ value.

robertbartel: The initial intent was for non-temporary datasets to not have an expire time, as this intuitively makes sense.  We could change how this is represented, though, if there is good reason.

python/lib/core/dmod/core/dataset.py
aaraney: Sorry, I should have been more specific. Should `category` and `data_domain` be Optional? In the previous `__init__`, they were required, non-optional, parameters, but in the previous implementation of `factory_init_from_deserialized_json` both _could_ be `None`.

python/lib/core/dmod/core/dataset.py
aaraney: I see 232 is unreachable if `self._expires` is `None`. I think I marked this as a TODO on a first quick pass and my bad for not double checking the logic in the `is_temporary`  property. Ill remove these comments in the source.

python/lib/core/dmod/core/dataset.py
robertbartel: Ah, I see. I believe that was a bug. It would not have made sense to have a dataset without a category or without a domain. I seem to have just left out the sanity check from the deserializer function.

aaraney: Great, thanks! I figured that might make for an interesting dataset.

robertbartel:
One thing I have noticed as I am reviewing the changes is that it _appears_ (correct me if I'm wrong) this will undo (pseudo) read-only characteristics of a large number of attributes across the _Serializable_ subtypes.   I.e., as if every `@property `x` implemented with a backing `_x` variable and without a `@x.setter` was having the setter added.  It overall reduces encapsulation in the design of these classes.  

I'm interested if I've missed something either being done or that could be done with **pydantic** that avoids this side effect.  

If there isn't a way around it, I don't want that to kill this idea.  But it will require a more deep consideration of design implications than I first thought.

aaraney:
> One thing I have noticed as I am reviewing the changes is that it _appears_ (correct me if I'm wrong) this will undo (pseudo) read-only characteristics of a large number of attributes across the _Serializable_ subtypes. I.e., as if every `@property `x`implemented with a backing`_x`variable and without a`@x.setter` was having the setter added. It overall reduces encapsulation in the design of these classes.
> 
> I'm interested if I've missed something either being done or that could be done with **pydantic** that avoids this side effect.

You are correct. In the current state, there is nothing stopping mutation through a property name (i.e. `foo.prop = 42`). `Pydantic` offers us two options around this using [model `Config` options](https://docs.pydantic.dev/usage/model_config/#options). Setting either the `allow_mutation` to `False` or `frozen` too `True` in [`Serializable`'s `Config` class](https://github.com/NOAA-OWP/DMOD/blob/61d80ea5771e3dfeb1c08b5735e2cffe3a54a91a/python/lib/core/dmod/core/serializable.py#L66) will change this behavior everywhere. Both of these options add faux-immutablity, meaning you can't set properties using `__setattr__`. The only difference is `frozen` makes subclasses hashable (`__hash__`).

In places where we allow mutation, I would advocate that we use explicit setters (`set_some_property()`). If that's not desirable, we can always toggle `allow_mutation` / `frozen` where it is needed.

robertbartel:
> Setting either `allow_mutation` to `False` or `frozen` too `True` in [Serializable's Config class](https://github.com/NOAA-OWP/DMOD/blob/61d80ea5771e3dfeb1c08b5735e2cffe3a54a91a/python/lib/core/dmod/core/serializable.py#L66) will change this behavior everywhere. Both of these options add faux-immutablity, meaning you can't set properties using `__setattr__`.

We are starting to go down a rabbit hole here though, drifting further from the initial goal of clearer, more maintainable code.  

Take the _DataRequirement_ class as an example.  It implements `__hash__` to not depend on all attributes, because not all its attributes are (faux) immutable.  This is necessary, too:  e.g., for `fulfilled_by`, it will not always be known how a requirement will be fulfilled at object creation time.  

It sounds like we'll have to implement classes to each have its own _Config_ to apply `allow_mutation=False`.  That still doesn't seem to properly addresses types like _DataRequirement_, which need mixed mutability.  I'm not sure I see how we can without adding another layer of abstraction for "un-mixing" the mutability.

aaraney:
> We are starting to go down a rabbit hole here though, drifting further from the initial goal of clearer, more maintainable code.
>
> Take the _DataRequirement_ class as an example. It implements `__hash__` to not depend on all attributes, because not all its attributes are (faux) immutable. This is necessary, too: e.g., for `fulfilled_by`, it will not always be known how a requirement will be fulfilled at object creation time.

I think that is fair criticism. I am not advocating that we use `frozen` (implements `__hash__`) for that reason. I should have been more clear about that. Likewise, I was going off the assumption that, in general, subtypes of `Serializable` are immutable (with some caveats; more about that later). In either case, I don't foresee `pydantic` adding anymore complexity to the areas of code you'd mentioned. Subtypes that need to implement, what I'll call, _partial_ hash can implement `__hash__` and types that are fully hashable can toggle `frozen=True` in the `Config` class for convenience. 

In the case of exposing internal properties, my thinking is, if we assume that `Serializable` subtypes are immutable, it is most clear to a user of that type that something _is_ mutable if that type exposes a setter method(s)(i.e. `set_status`). Otherwise, they can assume all properties disallow mutation. Now, how to make that happen gets to your point regarding maintainability. In the simple case if we assume `allow_mutation=False` on `Serializable` we could expose a setter like so:

```python
from dmod.core.serializable import Serializable

class Person(Serializable):
  name: str
  age: int

  def set_age(self, value: int) -> None:
    self.__dict__["age"] = value

```

I am not thrilled that we have to go through `__dict__`, but at the same time, does this use-case surface often enough to be overly concerned?

robertbartel:
> I am not thrilled that we have to go through __dict__, but at the same time, does this use-case surface often enough to be overly concerned?

Yeah, I could live with it, but I'm not crazy about it either.  There are maybe 5 or 6 classes (throughout DMOD, not just in _dmod-core_) to which this would apply.  The one that clashes the most is probably _Job_ and its subtypes, due to how often job objects have to be updated and how many places deal with that.

For now, let's stick with your original plan that keeps things mutable.  I think that's preferable to a switch to `set_x` functions in a lot of different places.  I'll open an issue to track improving the implementation to account for encapsulation better in the future.  

There is one more item holding this up:  the tests.  Those understandably fail because the scope of this PR was limited.  Before final approval and merging, I'd like to have the full sequence of necessary related PRs essentially ready to go.  Ideally, this will allow us to get to a final branch in the sequence where we see all the tests passing.  Then we can go through a relatively quick "merge-eldest, rebase-next, repeat" routine.  This will also minimize how long `master` is in a state where it produces incompatible packages.

robertbartel:
Actually, @aaraney, an example just occurred to me where the encapsulation is importation.  For _Job_ in particular, `last_updated` may prevent us from just sticking with your initial approach.  It is a mutable property, but not directly.  There is a function to reset it (i.e., to "now"), but it is otherwise modified by the setters of other properties.

As I read through the _Job_ classes now, a few more encapsulation issues jump out.  Have a look at `_allocation_service_names`, `status`, `status_phase`,  and `status_step` properties, which all have related behavior encapsulated in other properties' getter implementations.  These aren't simply addressed by the `__dict__`-based setter solution.

It looks like a third PR for _dmod-scheduler_ is going to need to be prepared and ready before we can approve this, although I am starting to wonder if we really need to combine the changes in all the package in a single PR.

aaraney:
> For now, let's stick with your original plan that keeps things mutable. I think that's preferable to a switch to `set_x` functions in a lot of different places. I'll open an issue to track improving the implementation to account for encapsulation better in the future.

Sounds like a plan. As always, thanks for looking at this critically.
 
> There is one more item holding this up: the tests. Those understandably fail because the scope of this PR was limited. Before final approval and merging, I'd like to have the full sequence of necessary related PRs essentially ready to go. Ideally, this will allow us to get to a final branch in the sequence where we see all the tests passing. Then we can go through a relatively quick "merge-eldest, rebase-next, repeat" routine. This will also minimize how long `master` is in a state where it produces incompatible packages.

and

> It looks like a third PR for _dmod-scheduler_ is going to need to be prepared and ready before we can approve this, although I am starting to wonder if we really need to combine the changes in all the package in a single PR.

Thanks for explicitly mentioning this. In retrospect, I think we will thank ourselves for keeping a ledger of decisions. 

Akin to your suggestion, I had in mind introducing these changes as a set of PRs. It might be advantageous of use to use a feature branch approach here. Meaning, create something like `pydantic-feature-branch` off this point in `master` and merge subsequent PRs that refactor particular libraries onto that feature branch. Once we are happy with the feature branch and _all_ libraries have been ported to `pydantic`, we can merge the feature branch onto `master` hopefully with more confidence.

aaraney:
> Actually, @aaraney, an example just occurred to me where the encapsulation is importation. For _Job_ in particular, `last_updated` may prevent us from just sticking with your initial approach. It is a mutable property, but not directly. There is a function to reset it (i.e., to "now"), but it is otherwise modified by the setters of other properties.
> 
> As I read through the _Job_ classes now, a few more encapsulation issues jump out. Have a look at `_allocation_service_names`, `status`, `status_phase`, and `status_step` properties, which all have related behavior encapsulated in other properties' getter implementations. ~These aren't simply addressed by the `__dict__`-based setter solution~ (actually, maybe they can be, though it would be extensive and feel messy).

Right, setters that modify other properties are easy if we could just transition everywhere to `set_x`, but that doesn't seem like a good idea at this stage. If it's safe to say that we will to move to `set_x` semantics in the future _and_ this only affects a small portion of the code base, we could do something like the following:

```python
from datetime import datetime
from pydantic import PrivateAttr
from dmod.core.serializable import Serializable

from typing import Callable, ClassVar, Dict

class Job(Serializable):
    id: str
    date: datetime = Field(default_factory=datetime.now)

    def _reset_date(self):
        self.date = datetime.now()

    def set_id(self, value: str):
        if value == self.id:
            return

        self._reset_date()
        self.id = value

    # map of field names to setters.
    # setters should be: Fn(self, value: T) -> None
    _setter_methods: ClassVar[Dict[str, Callable]] = {"id": set_id}

    _setattr_recursed: bool = PrivateAttr(False)

    def __setattr__(self, name, value):
        if name not in self._setter_methods:
            return super().__setattr__(name, value)

        if self._setattr_recursed:
            self.__dict__[name] = value
            self._setattr_recursed = False

        setter_fn = self._setter_methods[name]

        message = f"Setting by attribute is deprecated. Use `{setter_fn.__name__}` method instead."
        warn(message, DeprecationWarning)

        self._setattr_recursed = True
        setter_fn(self, value)
```

robertbartel:
@aaraney, per your suggestion, I've created a feature branch for this line of changes in the main upstream repo - `pydantic-serializable`.  I also went ahead and changed this PR to be based on that branch.

Once the conflict is resolved, we can approve, rebase, and merge this into that branch.

aaraney:
@robertbartel, okay it should be safe to approve and rebase this now.

aaraney:
Merge conflict forced me to reopen this work in another PR (#256). In that PR, the work is introduced into the repo, however conversation and description still live here.

PR #217, opened 2022-11-14
Add Ngen Calibration Request Variant

aaraney:
Add support for Ngen Calibration Requests. This request type is identical to [`NGENRequest`'s](https://github.com/NOAA-OWP/DMOD/blob/88dc3666ba06dfc782e2e8bcac21d69fccbc406c/python/lib/communication/dmod/communication/maas_request.py#L841) with the caveat of the `model_name` field which is now `ngen_cal` instead of `ngen`.

The `schedule` has been updated to handle the `NGENCalibrationRequest` variant. `NGENCalibrationRequest`s, for now, use the container image tag pair of `127.0.0.1:5000/ngen_cal:latest`.

## Additions

- Add support for Ngen Calibration Requests.
- Scheduler supports Ngen Calibration Requests. Expects container image tag pair `127.0.0.1:5000/ngen_cal:latest` to handle these requests.

## Changes

- Search for valid `ModelExecRequest` subtypes is now recursive.

## Todos

- Determine requirements and implement a `NGENCalibrationRequestResponse` variant.
- Create `ngen_cal` Dockerfile.
- [ ] bump `communication` and `scheduler` minor versions.

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows project standards (link if applicable)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #240, opened 2023-01-11
Combined pre-AGU GUI and Calibration Message Work

robertbartel:
This PR is a combination of the map view GUI work, the dataset management GUI work, and the calibration request message work done prior to AGU Fall Meeting 2022.  The original branches for that work have been merged so we don't have lingering, non-aligning, outstanding branches.  Also, other subsequent work has been rebased in.

We may still end up with a queue of branches, but if we do, this should at least help us ensure things happen in a series and are thus easier to maintain.

This PR effectively replaces #197, #210, and #217.


aaraney:
@robertbartel, looking at the files the PR affects, I am a little concerned how merging this will affect merging the pydantic refactor work. It might be painful, but we might try doing a dry run of that merge process locally before we do it on github just to see what the consequneces are. I guess a precursor that that thought is, how "complete" does the work this PR introduces need to be? What im really trying to get at is, what does the review process for the PR look like? Speaking for some of my commits in this PR, there are certainly some loose ends that should be tied up or just dropped and have new issues opened to address that work in the future.

robertbartel:
@aaraney, I performed a local rebase of the Pydantic changes in the combined branch you'd sent me, `refactor-libs`, across this branch and those queued/blocked in the same sequence (I may have left out the last one for now).  I may not have pushed them all, but I think I've at least pushed the last of these - `austin_refactor_w_postagu_4` - to my fork.  We can double check before things are committed, but I think the hard part there is done.

In terms of completeness, we expect things to be incomplete at this point.  We need to keep track of loose ends and broken things but not make this PR perfect, even for the things within its scope.  The reason for the associated PR sequence is that other work builds off this further to continue to iteratively address things.

aaraney:
> @aaraney, I performed a local rebase of the Pydantic changes in the combined branch you'd sent me, `refactor-libs`, across this branch and those queued/blocked in the same sequence (I may have left out the last one for now). I may not have pushed them all, but I think I've at least pushed the last of these - `austin_refactor_w_postagu_4` - to my fork. We can double check before things are committed, but I think the hard part there is done.

@robertbartel, right I am forgetting that you did that. I think the `refactor-libs` branch is basically what the `pydantic-refactor` branch looks like now, so it seems will just have some merge conflicts that we will have to carefully resolve.

> In terms of completeness, we expect things to be incomplete at this point. We need to keep track of loose ends and broken things but not make this PR perfect, even for the things within its scope. The reason for the associated PR sequence is that other work builds off this further to continue to iteratively address things.

Thanks for clarifying that. To push this PR forward, ill go ahead and open up TODO issues so we can circle back to incomplete code introduced by this PR.

aaraney:
The more I look at this, the more I am concerned about the amount of dead code and technical debt this PR will introduce. I'm not sure this is avoidable, but I thought I would voice my thoughts either way.

robertbartel:
Before we move forward and further with this (including fixing the current conflicts), I am going to spend some time looking at what it will take to untether and re-separate the distinct aspects, as well as the subsequent changes in the related series of PRs (#276, #277, #278, and #279).  

In most circumstances, I feel the technical debt and dead code @aaraney alluded to earlier would be essentially unavoidable:  a side effect of the development process for something this big when it is done in chunks.  That's even with this set of changes being bigger and including more than we normally want to see.

Here, though, with a fairly big shift in GUI code being planned in #286, the direct need for large parts of this (long-term at least) is now questionable.   It may be better at this point to cherry-pick certain things out, even if that comes with a bit of work to resolve resulting conflicts.

aaraney:
@robertbartel, what is the state of this? I've not done much digging, but it seems that the bulk of this work has merged through other PRs. Can we close this?

aaraney:
@robertbartel, is it fair to close this?

robertbartel:
@aaraney, I'm not quite ready to yet.  Given recent decisions to move away from a React-based GUI, it's not impossible we'd want to come back to some of this, and I'm not sure it's all been merged.  Let's wait for now.

PR #244, opened 2023-01-17
Fix bug with test sources not being excluded from Python packages

### LINKED ISSUE #241: Properly apply the incorrectly attempted exclusion of test sources from Python packages
As discussed in #229 and #207, test sources are currently being packaged when we create the DMOD Python packages, and this is causing problems.  It's also unnecessary.  In fact, the intent was to excluded them, but this was being done incorrectly.    

Most, if not all the Python packages have a line much like this in their _setup.py_ file; this example is from _python/lib/core/setup.py_:

``` python
packages=find_namespace_packages(exclude=('tests',` 'schemas', 'ssl', 'src'))
```

The directory structure looks like this:

```
python/lib/core
├── README.md
├── dmod
│   ├── core
│   └── test
└── setup.py
```

To actually exclude sources in _dmod/test_, the above line from _setup.py_ should read:

``` python
packages=find_namespace_packages(exclude=['dmod.test', 'schemas', 'ssl', 'src'])
```

This needs to be corrected for (or added to, if any were missed) all the DMOD namespace packages.

robertbartel:
Adjusting _setup.py_ for Python packages to properly exclude test sources.  Also bumping package version numbers.

Fixes #241.


aaraney (approved):
Looks good to me!

PR #276, opened 2023-02-13
Support for alternate data formats

robertbartel:
Adding support for data formats to have satisfactory alternate formats.  This enables, e.g., a requirement for a forcing dataset to be fulfilled by a dataset with a format that doesn't strictly match that specified by the requirement.

The branch for this commit is based on the branch in #240, making that a blocker for this.   It should not be reviewed until #240 is complete.


robertbartel:
Closing; replaced by #299, which isolates the relevant changes.

PR #299, opened 2023-03-14
Support for alternate data formats

robertbartel:
Adding support for data formats to have satisfactory alternate formats. This enables, e.g., a requirement for a forcing dataset to be fulfilled by a dataset with a format that doesn't strictly match that specified by the requirement.

Note that this supersedes #276.  It is effectively the same change set (small modifications were done due to package versions needing to be further bumped) rebased onto the current `master` commit.


aaraney (dismissed):
Looks good to me! All I would ask is to open an issue to track adding tests to this area of the code in the future!

robertbartel:
@aaraney, I've rebased and force-push the updated branch, so this will need another review.

aaraney (approved):

PR #278, opened 2023-02-13
Add support for serial ngen jobs

robertbartel:
- Update _NGENRequest_ to support request of either serial or parallel jobs
- Add support for performing serial NextGen execution to `ngen` Docker image entrypoint
- Update _Launcher_ scheduler class to support running serial NextGen jobs
- Fix bug in _Launcher_ related to handling of _NgenCalibrationRequest_
- Update both dataservice and partitionerservice to handle serial NextGen jobs properly if/when encountered, since these  don't need to ever await partitioning

The branch for this PR is based on the one for #277; as such, that PR blocks the review of this one.

PR #301, opened 2023-03-15
Add support for serial ngen jobs

robertbartel:
- Update _NGENRequest_ to support request of either serial or parallel jobs
- Add support for performing serial NextGen execution to `ngen` Docker image entrypoint
- Update _Launcher_ scheduler class to support running serial NextGen jobs
- Update both dataservice and partitionerservice to handle serial NextGen jobs properly if/when encountered, since these  don't need to ever await partitioning

Note that this supersedes #278. It is essentially the same change set rebased onto the current master commit.


aaraney (approved):
Looks good to me!

Issue #292, opened 2023-03-13
Update workflow logic at AWAITING_DATA step to make sure that required realization config datasets are created - derived from the instance of newly created job config class - and associated with job to fulfill requirement


robertbartel:

robertbartel:
Blocked by other, prerequisite subtasks of #251.

Issue #34, opened 2020-09-07
schduler-service package missing cryptography dependency in setup.py


hellkite500:
https://github.com/NOAA-OWP/DMOD/blob/85c62cfdcfac7075fad7aecf4d24d3ea68a8346c/python/services/schedulerservice/setup.py#L17

Need to add `cryptography` to the `install_requires` for this service package.

PR #316, opened 2023-03-29
PartialRealizationConfig class and usage in requests and dataservice

### LINKED ISSUE #263: Create new DMOD class, similar in purpose and composition to `NgenRealization` from `ngen.config` package, to represent only those job config details provided by user via GUI or CLI when requesting a job

robertbartel:
- Create new _PartialRealizationConfig_ class to represent partial realization config, based on details that need to be user-supplied
- Update _AbstractNgenRequest_ to utilize _PartialRealizationConfig_
- Update _dataservice.ServiceManager_ class to be able to both 
    - recognize and support situations when a realization config isn't available for a request job but can be derived because of an included _PartialRealizationConfig_ in the request
    - derived realization configs and containing datasets from a job request's _PartialRealizationConfig_ when needed

This closes #263, #291, and #292.


robertbartel:
@aaraney, I might need your help a bit with finishing the tests.

I'm attempting to use _NgenRealization_ objects from _ngen-config_ as a part of the unit test setup.  I initialize those from realization config files essentially copied from the _ngen_ repo.  Or at least I try to.  I have beat my head into a wall of validation errors for most of today.  For the first of the two files, they look like what I've copied below.

I eventually did a quick swap of the realization config from the AGU town hall example setup, and it did not fail like this.  But even with looking at the errors and comparing the two configs side by side, I can't tell what is causing these issues.

Some of the error are just unclear (the `global -> formulations -> 0 -> params -> __root__` items).  Others don't make sense, assuming I am interpreting them correctly (`unexpected value; permitted: 'bmi_python' (type=value_error.const; given=bmi_multi; permitted=['bmi_python'])`).  

```
Error
Traceback (most recent call last):
  File "/Users/rbartel/Developer/noaa/dmod/python/lib/communication/dmod/test/test_partial_realization_config.py", line 66, in setUp
    self._base_real_configs: List[NgenRealization] = self.read_example_realization_config_sources()
  File "/Users/rbartel/Developer/noaa/dmod/python/lib/communication/dmod/test/test_partial_realization_config.py", line 61, in read_example_realization_config_sources
    return [NgenRealization(**json.load(cfg_path.open())) for cfg_path in config_paths]
  File "/Users/rbartel/Developer/noaa/dmod/python/lib/communication/dmod/test/test_partial_realization_config.py", line 61, in <listcomp>
    return [NgenRealization(**json.load(cfg_path.open())) for cfg_path in config_paths]
  File "pydantic/main.py", line 342, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 48 validation errors for NgenRealization
global -> formulations -> 0 -> params -> __root__
  expected str, bytes or os.PathLike object, not NoneType (type=type_error)
global -> formulations -> 0 -> params -> __root__
  expected str, bytes or os.PathLike object, not NoneType (type=type_error)
global -> formulations -> 0 -> params -> __root__
  expected str, bytes or os.PathLike object, not NoneType (type=type_error)
global -> formulations -> 0 -> params -> __root__
  expected str, bytes or os.PathLike object, not NoneType (type=type_error)
global -> formulations -> 0 -> params -> name
  unexpected value; permitted: 'bmi_python' (type=value_error.const; given=bmi_multi; permitted=['bmi_python'])
global -> formulations -> 0 -> params -> main_output_variable
  unexpected value; permitted: 'land_surface_water__runoff_depth' (type=value_error.const; given=Q_OUT; permitted=('land_surface_water__runoff_depth',))
global -> formulations -> 0 -> params -> __root__
  expected str, bytes or os.PathLike object, not NoneType (type=type_error)
global -> formulations -> 0 -> params -> modules -> 0 -> params -> name
  unexpected value; permitted: 'bmi_c' (type=value_error.const; given=bmi_c++; permitted=['bmi_c'])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> model_type_name
  unexpected value; permitted: 'TOPMODEL' (type=value_error.const; given=bmi_c++_sloth; permitted=['TOPMODEL'])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> name
  unexpected value; permitted: 'bmi_c' (type=value_error.const; given=bmi_c++; permitted=['bmi_c'])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> model_type_name
  unexpected value; permitted: 'CFE' (type=value_error.const; given=bmi_c++_sloth; permitted=['CFE'])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> name
  unexpected value; permitted: 'bmi_c' (type=value_error.const; given=bmi_c++; permitted=['bmi_c'])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> main_output_variable
  unexpected value; permitted: 'water_potential_evaporation_flux' (type=value_error.const; given=z; permitted=('water_potential_evaporation_flux',))
global -> formulations -> 0 -> params -> modules -> 0 -> params -> name
  unexpected value; permitted: 'bmi_fortran' (type=value_error.const; given=bmi_c++; permitted=['bmi_fortran'])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> model_type_name
  unexpected value; permitted: 'NoahOWP' (type=value_error.const; given=bmi_c++_sloth; permitted=['NoahOWP'])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> name
  unexpected value; permitted: 'bmi_python' (type=value_error.const; given=bmi_c++; permitted=['bmi_python'])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> main_output_variable
  unexpected value; permitted: 'land_surface_water__runoff_depth' (type=value_error.const; given=z; permitted=('land_surface_water__runoff_depth',))
global -> formulations -> 0 -> params -> modules -> 0 -> params -> model_type_name
  unexpected value; permitted: 'SLOTH' (type=value_error.const; given=bmi_c++_sloth; permitted=['SLOTH'])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> name
  unexpected value; permitted: 'bmi_multi' (type=value_error.const; given=bmi_c++; permitted=['bmi_multi'])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> init_config
  unexpected value; permitted: '' (type=value_error.const; given=/dev/null; permitted=[''])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> model_params
  unexpected value; permitted: None (type=value_error.const; given={'sloth_ice_fraction_schaake(1,double,m,node)': '0.0', 'sloth_ice_fraction_xinan(1,double,1,node)': '0.0', 'sloth_smp(1,double,1,node)': '0.0'}; permitted=[None])
global -> formulations -> 0 -> params -> modules -> 0 -> params -> modules
  field required (type=value_error.missing)
global -> formulations -> 0 -> params -> modules -> 1 -> params -> name
  unexpected value; permitted: 'bmi_c' (type=value_error.const; given=bmi_fortran; permitted=['bmi_c'])
global -> formulations -> 0 -> params -> modules -> 1 -> params -> model_type_name
  unexpected value; permitted: 'TOPMODEL' (type=value_error.const; given=bmi_fortran_noahowp; permitted=['TOPMODEL'])
global -> formulations -> 0 -> params -> modules -> 1 -> params -> name
  unexpected value; permitted: 'bmi_c' (type=value_error.const; given=bmi_fortran; permitted=['bmi_c'])
global -> formulations -> 0 -> params -> modules -> 1 -> params -> model_type_name
  unexpected value; permitted: 'CFE' (type=value_error.const; given=bmi_fortran_noahowp; permitted=['CFE'])
global -> formulations -> 0 -> params -> modules -> 1 -> params -> name
  unexpected value; permitted: 'bmi_c' (type=value_error.const; given=bmi_fortran; permitted=['bmi_c'])
global -> formulations -> 0 -> params -> modules -> 1 -> params -> main_output_variable
  unexpected value; permitted: 'water_potential_evaporation_flux' (type=value_error.const; given=QINSUR; permitted=('water_potential_evaporation_flux',))
global -> formulations -> 0 -> params -> modules -> 1 -> params -> model_type_name
  unexpected value; permitted: 'NoahOWP' (type=value_error.const; given=bmi_fortran_noahowp; permitted=['NoahOWP'])
global -> formulations -> 0 -> params -> modules -> 1 -> params -> name
  unexpected value; permitted: 'bmi_python' (type=value_error.const; given=bmi_fortran; permitted=['bmi_python'])
global -> formulations -> 0 -> params -> modules -> 1 -> params -> main_output_variable
  unexpected value; permitted: 'land_surface_water__runoff_depth' (type=value_error.const; given=QINSUR; permitted=('land_surface_water__runoff_depth',))
global -> formulations -> 0 -> params -> modules -> 1 -> params -> name
  unexpected value; permitted: 'bmi_c++' (type=value_error.const; given=bmi_fortran; permitted=['bmi_c++'])
global -> formulations -> 0 -> params -> modules -> 1 -> params -> model_type_name
  unexpected value; permitted: 'SLOTH' (type=value_error.const; given=bmi_fortran_noahowp; permitted=['SLOTH'])
global -> formulations -> 0 -> params -> modules -> 1 -> params -> name
  unexpected value; permitted: 'bmi_multi' (type=value_error.const; given=bmi_fortran; permitted=['bmi_multi'])
global -> formulations -> 0 -> params -> modules -> 1 -> params -> init_config
  unexpected value; permitted: '' (type=value_error.const; given=./data/bmi/fortran/noah-owp-modular-init-{{id}}.namelist.input; permitted=[''])
global -> formulations -> 0 -> params -> modules -> 1 -> params -> modules
  field required (type=value_error.missing)
global -> formulations -> 0 -> params -> modules -> 2 -> params -> model_type_name
  unexpected value; permitted: 'TOPMODEL' (type=value_error.const; given=bmi_c_cfe; permitted=['TOPMODEL'])
global -> formulations -> 0 -> params -> modules -> 2 -> params -> model_type_name
  unexpected value; permitted: 'CFE' (type=value_error.const; given=bmi_c_cfe; permitted=['CFE'])
global -> formulations -> 0 -> params -> modules -> 2 -> params -> main_output_variable
  unexpected value; permitted: 'water_potential_evaporation_flux' (type=value_error.const; given=Q_OUT; permitted=('water_potential_evaporation_flux',))
global -> formulations -> 0 -> params -> modules -> 2 -> params -> name
  unexpected value; permitted: 'bmi_fortran' (type=value_error.const; given=bmi_c; permitted=['bmi_fortran'])
global -> formulations -> 0 -> params -> modules -> 2 -> params -> model_type_name
  unexpected value; permitted: 'NoahOWP' (type=value_error.const; given=bmi_c_cfe; permitted=['NoahOWP'])
global -> formulations -> 0 -> params -> modules -> 2 -> params -> name
  unexpected value; permitted: 'bmi_python' (type=value_error.const; given=bmi_c; permitted=['bmi_python'])
global -> formulations -> 0 -> params -> modules -> 2 -> params -> main_output_variable
  unexpected value; permitted: 'land_surface_water__runoff_depth' (type=value_error.const; given=Q_OUT; permitted=('land_surface_water__runoff_depth',))
global -> formulations -> 0 -> params -> modules -> 2 -> params -> name
  unexpected value; permitted: 'bmi_c++' (type=value_error.const; given=bmi_c; permitted=['bmi_c++'])
global -> formulations -> 0 -> params -> modules -> 2 -> params -> model_type_name
  unexpected value; permitted: 'SLOTH' (type=value_error.const; given=bmi_c_cfe; permitted=['SLOTH'])
global -> formulations -> 0 -> params -> modules -> 2 -> params -> name
  unexpected value; permitted: 'bmi_multi' (type=value_error.const; given=bmi_c; permitted=['bmi_multi'])
global -> formulations -> 0 -> params -> modules -> 2 -> params -> init_config
  unexpected value; permitted: '' (type=value_error.const; given=./data/bmi/c/cfe/{{id}}_bmi_config.ini; permitted=[''])
global -> formulations -> 0 -> params -> modules -> 2 -> params -> modules
  field required (type=value_error.missing)
```

aaraney:
@robertbartel, sorry that `pydantic` assaulted you like this. Im taking a look now!

aaraney:
This is the source issue in `ngen-cal` that documents the problem: https://github.com/NOAA-OWP/ngen-cal/issues/32.

robertbartel:
Ok, I will have a look at those. If I can, I'll try to do a review on #34 so it can be merged in.

robertbartel:
FYI, @aaraney, I tried to install the package from the branch in [ngen-cal#34](https://github.com/NOAA-OWP/ngen-cal/pull/34) to see if I could get the tests passing.  It looked promising at first, down to just a problem with ``SLOTH`` failing because that model type wasn't added yet.  But switching to a copy of that branch that I rebased, I get a new set of errors.  I think you talk about this problem some over in the PR.

```Python traceback
Traceback (most recent call last):
  File "/Users/rbartel/Library/Application Support/JetBrains/Toolbox/apps/IDEA-U/ch-0/231.8109.175/IntelliJ IDEA.app.plugins/python/helpers/pycharm/_jb_unittest_runner.py", line 38, in <module>
    sys.exit(main(argv=args, module=None, testRunner=unittestpy.TeamcityTestRunner,
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/unittest/main.py", line 100, in __init__
    self.parseArgs(argv)
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/unittest/main.py", line 147, in parseArgs
    self.createTests()
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/unittest/main.py", line 158, in createTests
    self.test = self.testLoader.loadTestsFromNames(self.testNames,
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/unittest/loader.py", line 220, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/unittest/loader.py", line 220, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/unittest/loader.py", line 154, in loadTestsFromName
    module = __import__(module_name)
  File "/Users/rbartel/Developer/noaa/dmod/python/lib/communication/dmod/test/test_partial_realization_config.py", line 4, in <module>
    from ..communication.maas_request.ngen.partial_realization_config import PartialRealizationConfig
  File "/Users/rbartel/Developer/noaa/dmod/python/lib/communication/dmod/communication/__init__.py", line 2, in <module>
    from .client import DataServiceClient, InternalServiceClient, ModelExecRequestClient, ExternalRequestClient, \
  File "/Users/rbartel/Developer/noaa/dmod/python/lib/communication/dmod/communication/client.py", line 15, in <module>
    from .maas_request import ExternalRequest, ExternalRequestResponse, ModelExecRequest, ModelExecRequestResponse, NWMRequest, \
  File "/Users/rbartel/Developer/noaa/dmod/python/lib/communication/dmod/communication/maas_request/__init__.py", line 15, in <module>
    from .ngen import AbstractNgenRequest, NGENRequest, NGENRequestResponse, NgenCalibrationRequest, \
  File "/Users/rbartel/Developer/noaa/dmod/python/lib/communication/dmod/communication/maas_request/ngen/__init__.py", line 1, in <module>
    from .abstract_nextgen_request import AbstractNgenRequest
  File "/Users/rbartel/Developer/noaa/dmod/python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py", line 1, in <module>
    from .partial_realization_config import PartialRealizationConfig
  File "/Users/rbartel/Developer/noaa/dmod/python/lib/communication/dmod/communication/maas_request/ngen/partial_realization_config.py", line 2, in <module>
    from ngen.config.formulation import Formulation
  File "/Users/rbartel/Developer/noaa/dmod/venv/lib/python3.8/site-packages/ngen/config/formulation.py", line 17, in <module>
    Formulation.update_forward_refs()
  File "pydantic/main.py", line 816, in pydantic.main.BaseModel.update_forward_refs
  File "pydantic/typing.py", line 553, in pydantic.typing.update_model_forward_refs
    '''Return a list containing n references to x.'''
  File "pydantic/typing.py", line 524, in pydantic.typing.update_field_forward_refs
    def __eq__(self, other):
  File "pydantic/fields.py", line 552, in pydantic.fields.ModelField.prepare
  File "pydantic/fields.py", line 666, in pydantic.fields.ModelField._type_analysis
  File "pydantic/fields.py", line 780, in pydantic.fields.ModelField.prepare_discriminated_union_sub_fields
  File "pydantic/utils.py", line 794, in pydantic.utils.get_discriminator_alias_and_values
pydantic.errors.ConfigError: Field 'model_name' of model 'PET' needs to be a `Literal`

Process finished with exit code 1
```

robertbartel:
FYI, I was able to locally work around the issue and make sure tests pass.  For now, the tests in the committed revision are skipped.  I've opened #317 to make sure they get put back in once the _ngen-config_ issues are sorted out.

aaraney:
Thanks for pinging me to review this! Ill set aside some time midday to review this.

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
```diff
@@ -150,24 +162,41 @@ def __init__(self,
...
         catchments : Optional[Union[Set[str], List[str]]]
             An optional collection of the catchment ids to narrow the geospatial domain, where the default of ``None``
             or an empty collection implies all catchments in the hydrofabric.
-        bmi_cfg_data_id : Optional[str]
-            The optioanl BMI init config ``data_id`` index, for identifying the particular BMI init config datasets
-            applicable to this request.
+        partition_cfg_data_id : Optional[str]
+            The optional ``data_id`` of an existing dataset containing a partitioning config that should be used in the
+            requested job.
+        formulations : Optional[PartialRealizationConfig]
```
aaraney, line 174: Is it worth creating a new class hierarchy to handle this instead of handling this switch case based on if you are provided a partial config or a concrete config? So a new type that takes in a `PartialRealizationConfig`. Some class then would use that information in that message to create a `AbstractNgenRequest` subtype.

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
```diff
@@ -150,24 +162,41 @@ def __init__(self,
         hydrofabric_data_id : str
             A data identifier for the hydrofabric, for distinguishing between different hydrofabrics that cover the same
             set of catchments and nexuses (i.e., the same sets of catchment and nexus ids).
+        bmi_cfg_data_id : str
```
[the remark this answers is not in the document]
aaraney: I now see the todo in [line 199](https://github.com/NOAA-OWP/DMOD/pull/316/files#diff-1296d443f3d8d2ed6c3654655471a28c4d5fa001ab5d8bf30c8ea7695a51b24fR199)

> # TODO: (#needs_issue) need to make BMI id(s) optional once BMI config generator tool is integrated in workflow

python/lib/communication/dmod/test/test_partial_realization_config.py
```diff
@@ -0,0 +1,240 @@
...
+        realization_config = list()
+        config_files_dir = cls.find_project_root() / "data" / "example_realization_configs"
+        if not config_files_dir.is_dir():
+            msg = "Expected dir with example realization configs for {} setup not found at {}"
+            raise RuntimeError(msg.format(cls.__name__, config_files_dir))
+
+        # See docstring for the basic makeup of each of the contained configs
+        files = ["ex_realization_config_01.json", "ex_realization_config_02.json"]
+        for p in [config_files_dir.joinpath(f) for f in files]:
+            with open(p) as config_data:
```
aaraney, line 62: As an aside here, `pydantic.BaseModel` has a `parse_file()` `classmethod` that will read and parse a json file and return a validated model. Not that it matters here, but it should actually be faster than the standard `json` lib too which is kind of neat.

python/lib/communication/setup.py
```diff
@@ -22,6 +22,6 @@
     license='',
     include_package_data=True,
     #install_requires=['websockets', 'jsonschema'],vi
-    install_requires=['dmod-core>=0.1.2', 'websockets>=8.1', 'jsonschema', 'redis'],
+    install_requires=['dmod-core>=0.1.2', 'websockets>=8.1', 'jsonschema', 'redis', 'ngen-config>=0.1.1'],
```
aaraney, line 25: For now we should probably install from `ngen_config @ git+https://github.com/noaa-owp/ngen-cal@master#egg=ngen_config&subdirectory=python/ngen_conf`. If `ngen-config` ever gets on `pypi` we should switch it however.

python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -489,6 +495,96 @@ async def _async_process_query(self, message: DatasetManagementMessage) -> Datas
...
+        if forcing_dataset.data_format == DataFormat.NETCDF_FORCING_CANONICAL:
+            forcing_cfg_params['provider'] = 'NetCDF'
+        elif forcing_dataset.data_format == DataFormat.AORC_CSV:
+            forcing_cfg_params['provider'] = 'CsvPerFeature'
+
+        # TODO: (#needs_issue) introduce logic to examine forcing dataset and intelligently assess what the file
+        #  name(s)/pattern(s) should be if they aren't explicitly provided
+
+        if request.formulation_configs is not None and request.formulation_configs.forcing_file_pattern is not None:
+            forcing_cfg_params['file_pattern'] = request.formulation_configs.forcing_file_pattern
```
aaraney, line 541: Do we need a default else case for now?

python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -489,6 +495,96 @@ async def _async_process_query(self, message: DatasetManagementMessage) -> Datas
...
+
+        Returns
+        -------
+        Forcing
+            Forcing config object to be used in building a ngen realization config to satisfy this request.
+        """
+        # A user may work around dataset performance issues by mounting something directly from the host into
+        # containers, using an env var for the mount source (see code in scheduler.py).  We can support this by
+        # expecting a certain prefix (from_env:::) for either the file pattern or the file basename.
+        use_from_env_workaround = False
```
aaraney, line 520: Was this mistakenly left in? I assume this was added before it was refactored into `PartialRealizationConfig`?

python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -489,6 +495,96 @@ async def _async_process_query(self, message: DatasetManagementMessage) -> Datas
...
+            forcing_cfg_params['path'] += 'from_env'
+        else:
+            forcing_cfg_params['path'] += '{}/{}/'.format(DataCategory.FORCING.name.lower(), forcing_dataset_name)
+
+        if request.formulation_configs is not None and request.formulation_configs.forcing_file_name is not None:
+            forcing_cfg_params['path'] += request.formulation_configs.forcing_file_name
+
+        return Forcing(**forcing_cfg_params)
+
+    def _build_ngen_realization_config_from_request(self, request: AbstractNgenRequest, job: Job) -> NgenRealization:
```
aaraney, line 556: This looks good! Is is appropriate to leave a TODO here for generating model `init_config` datasets in the future?

aaraney: maybe this makes more sense in `_derive_realization_config_from_formulations`.

python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -716,23 +868,107 @@ async def can_be_fulfilled(self, requirement: DataRequirement) -> Tuple[bool, Op
...
+            Raised if a job with the correct status has an initially unfulfilled requirement for which a satisfactory
+            dataset can not be derived by this function.
+        """
+        # Only do something if the job has the right status
+        if job.status_step != JobExecStep.AWAITING_DATA:
+            return []
+
+        results = []
+
+        for req in [r for r in job.data_requirements if r.fulfilled_by is None]:
```
aaraney, line 960: totally unnecessary, but you could use a generator comprehension here instead of a list comprehension :).

python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -716,23 +868,107 @@ async def can_be_fulfilled(self, requirement: DataRequirement) -> Tuple[bool, Op
...
+            # **********************************************************************************************************
+            # *** NOTE: if/when deriving forcing datasets is supported, make sure this is done before config datasets
+            # **********************************************************************************************************
+            # Derive realization config datasets from formulations in message body when necessary
+            if req.category == DataCategory.CONFIG and req.domain.data_format == DataFormat.NGEN_REALIZATION_CONFIG:
+                self._derive_realization_config_from_formulations(requirement=req, job=job)
+                results.append(req)
+            # The above are the only supported derivations, so blow up here if there was something else
+            else:
+                msg_template = "Unsupported requirement dataset derivation for job {} (requirement: {})"
```
aaraney, line 970: It seems like we either want to batch create or roll back here. I am assuming that it is possible to successfully create a dataset in a prior iteration and then fail in the next. Maybe its just a difference of expectations, but I would expect that the prior created datasets would be removed. Thoughts?

python/lib/communication/setup.py
aaraney: I see that it is in `requirements.txt`, but you know im all ornery about being able to install things using just `pip` and `setup.py` / `setup.cfg` 😄.

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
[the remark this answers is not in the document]
robertbartel: > Are there multiple `init_config` files in a `bmi_cfg_dataset`?

There certainly can be, yes, and I expect that to be the typical case.  There is no strict requirement or guarantee for this, though.

> Related and sort of an aside, I am starting to wonder if simplifying and unifying the concept of a config_dataset to contain not only the `ngen_realization_config` but also the associated` bmi_cfg_dataset`s (`init_config` files), `partition_config`, and `routing_config` is worth it? This implies that the other dataset types would be dropped too. That would mean we would still validate / generate the same number of entities, but everything would live in a single `dataset`. This would be a big change and would need quite a bit of conversation, but as things scale out, the number of pointers, links, or references (whatever you want to call them) seem to be adding a fair amount of complexity and points of failure.

It's worth giving some thought.  One way or another, jobs need those artifacts (except for times when they don't), so I don't know how much we can truly clean up in the way of those associations.  Much of what we'd remove in links between things would  be replace by new requirements for validating whether a config dataset is "complete."  But leaning on something else to encapsulate the requirements and relationships may be still be better.

Regardless of what we do, I'm pretty confident `partition_config`s will need to remain separated.  Those are different in how they affects things.  For example, maybe someone wants to test different partitioning schemes for the same configuration, or compare parallel and serial execution.

aaraney (commented):
In general I think this looks really solid! Nice stuff! I have a few architectural questions that merit a little bit of discussion and potential changes, so for now I am going to leave this as a comment review. Once we've talked through those comments, I am happy to approve and merge.

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
robertbartel: In some ways I think that would be better.  I worry/thought it would not fit together well in certain places:  for example, the logic that is run during the `AWAITING_DATA_CHECK` and `AWAITING_DATA` job stages.

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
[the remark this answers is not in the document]
aaraney: > > Are there multiple `init_config` files in a `bmi_cfg_dataset`?
> 
> There certainly can be, yes, and I expect that to be the typical case. There is no strict requirement or guarantee for this, though.

Okay, and right, there isnt a strict requirement or guarantee that all `init_config` files in a `bmi_cfg_dataset` are for a single formulation? So for example, there could be an `init_config` dataset that contains configurations for `cfe` and `topmodel` formulations, in the case of possibly using a bmi multi formulation with `cfe` and `topmodel`.

> It's worth giving some thought. One way or another, jobs need those artifacts (except for times when they don't), so I don't know how much we can truly clean up in the way of those associations. Much of what we'd remove in links between things would be replace by new requirements for validating whether a config dataset is "complete." But leaning on something else to encapsulate the requirements and relationships may be still be better.

Right, this would move the associations into a single dataset, rather than having the associations spread across multiple datasets. I think we will gain a lot when it comes to reproducibility among other things, but we definitely need to weigh the pros and cons. There is a ton of history and thought that I know you've given to this problem that I want to better understand. We should schedule some time to talk about this maybe next week? 

> Regardless of what we do, I'm pretty confident `partition_configs` will need to remain separated. Those are different in how they affects things. For example, maybe someone wants to test different partitioning schemes for the same configuration, or compare parallel and serial execution.

We are in agreement here. Your examples sound like use-cases we will see by the end of the year. With that in mind, I am curious if some kind of configuration service isn't worth discussing. I started to allude to that use case in the next comment:

> Is it worth creating a new class hierarchy ...`AbstractNgenRequest` subtype.

Or maybe it is still in the scope of the `dataservice`?

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
aaraney: Right, I guess in the short term to get around the job semantics you could have a job that schedules another job. The first job really just determines data requirements for the second. Alternatively, it might be easier to roll this functionality into its own service like I mentioned above.

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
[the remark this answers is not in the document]
robertbartel: > Okay, and right, there isnt a strict requirement or guarantee that all init_config files in a bmi_cfg_dataset are for a single formulation?

Correct:  for the moment, a dataset in the `BMI_CONFIG` _DataFormat_, which is what we'd expect for all BMI init config datasets, can only have constraints enforced on a `GLOBAL_CHECKSUM` index.  The intent is for that to be an MD5 (I think) sum on all the contained files (it may actually be file names, rather than their contents).  

In other words, right now, a BMI dataset can essentially contain anything.  DMOD is able to imprecisely constrain things to check whether a dataset has exactly everything the user expected, but the details of that are wide open.  We definitely will need to refine this in the future to some degree, though.

python/services/dataservice/dmod/dataservice/service.py
robertbartel: I've added a few TODO comments about making sure this gets done, including one in a different spot where I already made a similar comment about the order of things if deriving a forcing dataset is needed.

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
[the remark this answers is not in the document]
robertbartel: > Right, this would move the associations into a single dataset, rather than having the associations spread across multiple datasets. I think we will gain a lot when it comes to reproducibility among other things, but we definitely need to weigh the pros and cons.

The biggest issue I see with this is that these are distinct, loosely coupled items.  A user may apply the same realization config in different experiments that use different sets of BMI init configs.  A user may also apply the same collection of BMI init configs to different experiments that use different realization configs.  And that's before we start to consider DMOD experiments using _multiple_ BMI init config datasets, which we can't support yet but have discussed. 

But that is also a very conceptual, perhaps too abstract viewpoint.  If we can adopt more practical approaches as pieces actually come together, and they don't block something we know that we need, I'm all for it.

python/services/dataservice/dmod/dataservice/service.py
robertbartel: The idea I had in mind was for some datasets to be temporary with an expiration time.  The data service would periodically clean up these.

I expected that some derived datasets may not be trivial to reproduce - derived forcings, for example - so we might want to hang onto them a little while after initially appearing to be done with them.  It also seemed likely that any datasets that were trivial to reproduce would also be trivial to store for a bit.

Of course, things are not fully implemented to make that work, but they've been started, and I added some more TODO comments around this area to help with that.

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
robertbartel: > Right, I guess in the short term to get around the job semantics you could have a job that schedules another job. The first job really just determines data requirements for the second. Alternatively, it might be easier to roll this functionality into its own service like I mentioned above.

I am concerned that will obfuscate things for the user.  We want/need to be able to start their job and give them back a job id right away, which they can use later to figure out how to get their output (or get back to a status page).  

We could still do something similar, albeit more complicated (e.g., have the initial job _spawn_ a sub-job for deriving data), but that is a larger scope that probably needs it's own future issue(s).

python/services/dataservice/dmod/dataservice/service.py
robertbartel: I didn't consider this and am not terribly skilled with my use of generators, so the discussion below is meant for learning purposes, not justification of the previous approach.

Would that save on anything other than the memory needed for the list object and it's references (small considering the referenced objects already exist)?  I don't know off the top of my head if, behind the scenes, the way `for` loop implementations work get something out of a generator here instead, although it seems to be just lazy versus eager application of the filtering.

aaraney: > Would that save on anything other than the memory needed for the list object and it's references (small considering the referenced objects already exist)?


Yes, but first a little setup:

We have:

- `n` items where `n == len(job.data_requirements)`
- `k` filtered items where `r.fulfilled_by is None`

So in the case of using the list comprehension you would do `n` iterations that result in a new list of `k` length. Then loop over the new list `k` times. So the total runtime is `Θ(n+k)`. By using the generator, you don't have to do an extra `k` iterations. So, the runtime is `Θ(n)`.

python/services/dataservice/dmod/dataservice/service.py
aaraney: That makes a lot of sense. I wasnt thinking about this in the context of long running tasks, like deriving forcing. Thanks for sharing your thoughts.

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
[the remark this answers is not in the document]
aaraney: Given your work, @robertbartel, on composite datatypes, I think this and other related conversations will come out of that work. I think its safe to resolve this for now.

python/lib/communication/dmod/communication/maas_request/ngen/partial_realization_config.py
```diff
@@ -0,0 +1,68 @@
...
+    the scheduling logic, but the generated realization config must properly reflect it.
+
+    The workaround is triggered by adding specific prefix (::attribute:`_FROM_ENV_PREFIX`) and delimiter
+    (::attribute:`_FROM_ENV_DELIMIT`) substrings to either of the forcing config fields.
+    """
+
+    _FROM_ENV_DELIMIT: ClassVar[str] = ':::'
+    _FROM_ENV_PREFIX: ClassVar[str] = 'from_env'
+
+    hydrofabric_uid: str
```
aaraney, line 27: Should we consider enforcing this as a `UUID` type?

python/lib/communication/dmod/communication/maas_request/ngen/partial_realization_config.py
```diff
@@ -0,0 +1,68 @@
...
+    forcing_file_pattern: Optional[str] = None
+    """ Optional catchment-id-based pattern string for basename of per-catchment forcing files. """
+
+    forcing_file_name: Optional[str] = None
+    """ Optional fixed name for the forcing data file. """
+
+    routing_config: Optional[Routing] = None
+    """ Optional routing config object for the partial config. """
+
+    is_env_workaround: bool = None
```
aaraney, line 45: `None` or `False`?

python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -489,6 +495,91 @@ async def _async_process_query(self, message: DatasetManagementMessage) -> Datas
...
+        Returns
+        -------
+        Forcing
+            Forcing config object to be used in building a ngen realization config to satisfy this request.
+        """
+        forcing_cfg_params = dict()
+
+        # Get the correct forcing dataset from associated requirement
+        # TODO: double check that this is being added when we do data checks
+        forcing_req = [r for r in request.data_requirements if r.category == DataCategory.FORCING][0]
```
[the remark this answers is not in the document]
aaraney: Ideally if that is an invariant that we are capturing in the concept of an `AbstractNgenRequest` we would enforce that there, so I guess I am more or less asking if that is an invariant? Or is it acceptable to have more than one forcing requirement?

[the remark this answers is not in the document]
aaraney: This is not a huge deal in the slightest, so if this is just a can of worms that im opening, please say so!

aaraney (approved):
I left a few minor comments that dont block this. Merge at will.

python/services/dataservice/dmod/dataservice/service.py
robertbartel: Ah, gotcha  Thanks for the explanation.  The PR has already been approved at this point, so I'm not going to change things now and mess that up, but I'll keep this in mind in the future.

python/lib/communication/dmod/communication/maas_request/ngen/partial_realization_config.py
robertbartel: The `hydrofabric_uid` is the id value generated by the hydrofabric itself.  It needs to be unique and _deterministic_:  essentially a hash value for the hydrofabric's data and metadata.  

I suppose we still _could_ use _UUID_ for it.  It just seems a little more tricky and a little less beneficial, given we aren't generating a random unique value.

python/lib/communication/dmod/communication/maas_request/ngen/partial_realization_config.py
robertbartel: I think we actually do want `None` here as a hint that other things are in place to make sure the value is always a _bool_.  The `default_is_env_workaround` validator function should make sure that _something_ other than `None` is always provided for `is_env_workaround`.

That's a style thing, though, so I'd be open to changing it in the future if we need.

python/services/dataservice/dmod/dataservice/service.py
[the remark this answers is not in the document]
robertbartel: _Probably_ ... but I'm not sure.  My biggest concerns at the moment related to multi-level hydrofabrics.  Will it make sense to keep distinct forcing datasets for individual layers?  Even if we end up still using a unified dataset, will we want/need to keep requirements separate per layer as we assemble those for a job request?  

I'd rather kick this can down the road for now.

python/lib/communication/dmod/communication/maas_request/ngen/partial_realization_config.py
aaraney: > I think we actually do want `None` here as a hint that other things are in place to make sure the value is always a _bool_.

We should check that. I think because you used `always=True` and `pre=True` this can result in `None`.

python/services/dataservice/dmod/dataservice/service.py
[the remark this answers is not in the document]
aaraney: > Even if we end up still using a unified dataset, will we want/need to keep requirements separate per layer as we assemble those for a job request?

I'd not thought about it in that way. For practical reasons, more or less for evaluation purposes, I'd assumed forcing products would be pre-processed into a single dataset, but that may be a poor assumption.

Issue #324, opened 2023-04-20
Pandas likely 2.0.0 causing modeldata test to fail


aaraney:
The `dmod.test.test_geopackage_hydrofabric.TestGeoPackageHydrofabric.test_uid_1_a` test is currently failing in several PRs. Below is a snipped from an action log showing the failure. 

[source](https://github.com/NOAA-OWP/DMOD/actions/runs/4725463216/jobs/8384466740#step:10:189)
```
===========================================================================
............................F...............
======================================================================
FAIL: test_uid_1_a (dmod.test.test_geopackage_hydrofabric.TestGeoPackageHydrofabric)
Test that the hydrofabric instance for example one has the expected unique id.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/DMOD/DMOD/python/lib/modeldata/dmod/test/test_geopackage_hydrofabric.py", line 309, in test_uid_1_a
    self.assertEqual(hydrofabric.uid, expected_uid)
AssertionError: '7b022f401ea2da1fdce2c1c2e36a8664b2299778' != '8a24b5eeae2596ceaf21058c49a27c8ae6f444ab'
- 7b022f401ea2da1fdce2c1c2e36a8664b2299778
+ 8a24b5eeae2596ceaf21058c49a27c8ae6f444ab
```

I compared the dependency versions installed when the tests were passing with the failing tests and it seems that `pandas==2.0.0` is the likely culprit. The last known pandas version that works is `1.5.3`. I tested this locally with `fiona` version `1.9.1` and `1.9.3` and `pandas==1.5.3` and the tests passed. However, there is one [outlier action](https://github.com/NOAA-OWP/DMOD/actions/runs/4658765370/jobs/8244825421#step:5:73) with `pandas==2.0.0` and `fiona==1.9.2` installed that passed? Im still a little puzzled about that one and ive not been able to reproduce it locally (yet, ill do that in the morning, there isn't a `fiona` wheel for that version for my machine). 

[Passing with `pandas==1.5.3`](https://github.com/NOAA-OWP/DMOD/pull/296/checks#step:5:74)
[Failing with `pandas==2.0.0`](https://github.com/NOAA-OWP/DMOD/actions/runs/4725463216/jobs/8384466740#step:5:80)
[Weird passing test `pandas==2.0.0`](https://github.com/NOAA-OWP/DMOD/actions/runs/4658765370/jobs/8244825421#step:5:73)

[...]

PR #326, opened 2023-04-24
Address issues with Geopackage hydrofabric unique id tests

robertbartel:
Changes implementation for generating unique id within _GeoPackageHydrofabric_ class to be more resilient to changes, the lack of which contributed to #324.  Also modified associated unit test to account for differences.


aaraney (approved):

closes #324

Issue #330, opened 2023-05-24
Intended behavior of is_integer and is_float


aaraney:
While working on #314, it was not clear to me the intended behavior of `is_integer` and `is_float`.

https://github.com/NOAA-OWP/DMOD/blob/7cab36fe792ccfc8d9f679ec71bc432fed568485/python/lib/core/dmod/core/common/helper_functions.py#L29-L54

My inclination is that `is_integer(42)` and `is_float(42.0)` should both return `True` but both return `False`. @christophertubbs, is this the intended behavior?

aaraney:
My best guess is that these functions could be re-written as:

```python
def is_integer(value: Any) -> bool:
  return isinstance(value, (int, np.integer))

def is_float(value: Any) -> bool:
  return isinstance(value, (float, np.floating))
```

However since `dmod.core` does not depend on `numpy` they can't be written that way. My guess is that why they are written how they are now. That leads me to think they should live in a module that depends on `numpy`. Thoughts @christophertubbs?

christophertubbs:
Found the issue - `value` at the end of each function is a `type`, so it'll _never_ be an instance of `float` or `int`. Those will need to be rewritten as something like:

```python
def is_integer(value) -> bool:
    value_type = value if type(value) == type else type(value)
  
    mro = get_mro_names(value_type) 
  
    module_name = value_type.__module__ if hasattr(value_type, "__module__") else '' 
  
    if module_name.startswith("numpy"): 
        return "integer" in mro
     
    loaded_integer_types = [int] + int.__subclasses__()
  
    return value in loaded_integer_types if type(value) == type else  isinstance(value, int)
```

aaraney:
> those returning false is _definitely_ wrong. The idea was that this would be usable in any library, whether it used numpy or not and requiring specialized logic if so because the mro for numpy types can be all types of funky.

Yeah, my only reservation is that to properly test these functions we would need to depend on `numpy` as a dev dep and that _feels_ wrong. Like we are solving a problem that we don't have in this context. Do you think that is fair? Or am I being overly harsh?

aaraney:
> Found the issue - `value` at the end of each function is a `type`, so it'll _never_ be an instance of `float` or `int`. Those will need to be rewritten as something like:
> 
> ```python
> def is_integer(value) -> bool:
>     value_type = value if type(value) == type else type(value)
>   
>     mro = get_mro_names(value_type) 
>   
[...]

PR #333, opened 2023-05-30
`get_common_type` fixes

christophertubbs:
Fixed some of the logic for determining if iterables have a common type and what that type might be.

Addresses some of the concerns in #330

### DIFF
--- a/python/lib/core/dmod/core/common/helper_functions.py
+++ b/python/lib/core/dmod/core/common/helper_functions.py
@@ -27,31 +27,33 @@ def get_mro_names(value) -> list[str]:
 
 
 def is_integer(value) -> bool:
-    if type(value) != type:
-        value = value.__class__
+    value_type = value.__class__
 
-    mro = get_mro_names(value)
+    mro = get_mro_names(value_type)
 
-    module_name = value.__module__ if hasattr(value, "__module__") else ''
+    module_name = value_type.__module__ if hasattr(value_type, "__module__") else ''
 
     if module_name.startswith("numpy"):
         return "integer" in mro
 
-    return isinstance(value, int)
+    loaded_integer_types = [int] + int.__subclasses__()
+
+    return value in loaded_integer_types if type(value) == type else isinstance(value, int)
 
 
 def is_float(value) -> bool:
-    if type(value) != type:
-        value = value.__class__
+    value_type = value.__class__
 
-    mro = get_mro_names(value)
+    mro = get_mro_names(value_type)
 
-    module_name = value.__module__ if hasattr(value, "__module__") else ''
+    module_name = value_type.__module__ if hasattr(value_type, "__module__") else ''
 
     if module_name.startswith("numpy"):
         return "floating" in mro
 
-    return isinstance(value, float)
+    loaded_float_types = [float] + float.__subclasses__()
+
+    return value in loaded_float_types if type(value) == type else isinstance(value, float)
 
 
 _PRIMITIVE_TYPE_IDENTIFIERS = [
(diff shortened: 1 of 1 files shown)


python/lib/core/dmod/core/common/helper_functions.py
```diff
@@ -27,31 +32,31 @@ def get_mro_names(value) -> list[str]:
...
-    mro = get_mro_names(value)
+    if numpy and isinstance(value, numpy.integer):
+        return True
 
-    module_name = value.__module__ if hasattr(value, "__module__") else ''
+    loaded_integer_types = [int] + get_subclasses(int)
 
-    if module_name.startswith("numpy"):
-        return "integer" in mro
+    if numpy:
```
aaraney, line 40: I might be wrong, but I dont think this is necessary. `isinstance` is inclusive of subtypes.

python/lib/core/dmod/core/common/helper_functions.py
```diff
@@ -27,31 +32,31 @@ def get_mro_names(value) -> list[str]:
...
+        loaded_integer_types.extend(
+            [numpy.integer] + get_subclasses(numpy.integer)
+        )
 
-    return isinstance(value, int)
+    return value in loaded_integer_types if type(value) == type else isinstance(value, int)
```
aaraney, line 45: I think this function can be reduced to:

`return isinstance(value, int) if numpy is None else isinstance(value, (int, np.integer))`

python/lib/core/dmod/core/common/helper_functions.py
```diff
@@ -27,31 +32,31 @@ def get_mro_names(value) -> list[str]:
...
 
-    if module_name.startswith("numpy"):
-        return "floating" in mro
+    if numpy:
+        loaded_float_types.extend(
+            [numpy.floating] + get_subclasses(numpy.floating)
+        )
 
-    return isinstance(value, float)
+    return value in loaded_float_types if type(value) == type else isinstance(value, float)
```
aaraney, line 59: Like the `is_integer` function, I think this can be reduced to 
`return isinstance(value, float) if numpy is None else isinstance(value, (int, np.floating))`

aaraney (approved):
I think this can be simplified and refactored a bit, but Ill leave that to your discretion @christophertubbs. Thanks!

python/lib/core/dmod/core/common/helper_functions.py
christophertubbs: It is, but most of the numpy types don't inherit from the vanilla types. The numpy integer mro is: `(<class 'numpy.int32'>, <class 'numpy.signedinteger'>, <class 'numpy.integer'>, <class 'numpy.number'>, <class 'numpy.generic'>, <class 'object'>)`, which doesn't include `int`, so `numpy.int32(9).__class__ in [int] + int.__subclasses__()` is `False`.

python/lib/core/dmod/core/common/helper_functions.py
christophertubbs: Only reason not to do that is to cover the `is_integer(int)` case you presented in the other ticket.

python/lib/core/dmod/core/common/helper_functions.py
christophertubbs: Same deal - if we're to cover `is_float(float)`, then we'll need the additional `loaded_float_types` stuff.

aaraney (approved):
Looks good to me! Thanks @christophertubbs. 

Closes #330.

PR #353, opened 2023-06-09
Improve handling of blosc2 and tables installation in ngen Docker image build

### LINKED ISSUE #312: Address ngen Docker image build issue with Python tables and blosc2 packages
The _ngen_ Docker image is failing to build due to a problem with the _tables_ Python dependency.  In the `rocky-ngen-deps` build stage, there are these two lines (probably 349).  The second triggers the error, but the first ends up being directly related.

``` Dockerfile
    && pip3 install numpy pandas pyyaml bmipy Cython blosc2 netCDF4 wheel packaging \
    && HDF5_DIR=/usr pip3 install --install-option="\'--hdf5 \${HDF5_DIR}\'" --install-option="\'--jobs=$(nproc) \'"  --no-build-isolation tables \
```

The error (shown below) complains about _skbuild_, but the root of the problem is that _tables_ requires a specific version of _blosc2_, which _at present_ is different than the version of _blosc2_ installed by the previous command.  This then causes dependency issues, and eventually the _tables_ install fails.

This can be fixed by specifying `blosc2==2.0.0` in the first command, but this is a brittle solution.  One way or another, this needs to be fixed, but a more robust solution would be preferable if one is feasible.

The error looks like this:

Screenshot 2023-03-06 at 14 17 42

robertbartel:
Update to _ngen_ Dockerfile related to _blosc2_ and _tables_ dependencies.

- Adds environment-based configuration of the version of _blosc2_ installed during image build
  - Includes fall-back to reasonable default version value when nothing is deliberately configured
- Removes use of `--install-option` pip flag in _tables_ install command, as this flag has been removed

This issue closes #312.

### DIFF
--- a/docker/main/ngen/Dockerfile
+++ b/docker/main/ngen/Dockerfile
@@ -48,6 +48,7 @@ ARG BOOST_VERSION=1.72.0
 ARG MPICH_VERSION="3.2"
 ARG MIN_PYTHON="3.8.0"
 ARG MIN_NUMPY="1.18.0"
+ARG BLOSC2_VERSION
 
 ARG NETCDF_C_VERSION=4.8.1
 ARG NETCDF_CXX_VERSION=4.3.1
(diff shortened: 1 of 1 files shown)


aaraney (approved):
This looks good to me. I trust that you've built the image to verify that it works? It appears `tables`'s [`blosc2`](https://github.com/PyTables/PyTables/blob/f8ad2f1e20bbb211ef40a8fa5423b69387115035/pyproject.toml#L13) dependency issues will be resolved in a future release. We should revisit this to see if we can remove this logic once `tables`'s `>3.8.0` has dropped.

PR #361, opened 2023-06-14
Fix for double logging

christophertubbs:
I found an issue where using the application specific loggers would log twice - once for the application logger and once through root. Removing the root configuration in the default setup limits the logging so that it is only logged once.


aaraney (approved):
Looks good to me, @christophertubbs. Thanks for fixing this!

PR #365, opened 2023-06-30
Pin pydantic<2

aaraney:
[Pydantic 2](https://pypi.org/project/pydantic/2.0/) was released today. This major version release introduces some breaking changes that we _may_ need to address in the future.

## Changes

- pin `pydantic>=1.10.8,~=1.10`

### DIFF
--- a/python/lib/communication/setup.py
+++ b/python/lib/communication/setup.py
@@ -21,7 +21,7 @@ setup(
     url='',
     license='',
     include_package_data=True,
-    install_requires=['dmod-core>=0.4.2', 'websockets>=8.1', 'jsonschema', 'redis', 'pydantic',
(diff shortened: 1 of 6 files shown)


PhilMiller:
It's generally preferable to specify a positive version constraint, like `pydantic~=1.10` ( think that's what we want here), so that it's explicit what *is* compatible rather than what isn't

aaraney:
@PhilMiller, that's fair. Thanks for the insight. I'll searched for the minimum supported version of `pydantic` and update this PR.

aaraney:
This is ready now.

robertbartel (approved):

aaraney:
Thanks, @robertbartel!

PR #367, opened 2023-07-05
Add data format value for t-route configs

robertbartel:
Creating a simple `DataFormat` value for the format of _t-route_ config data.

### DIFF
--- a/python/lib/core/dmod/core/meta_data.py
+++ b/python/lib/core/dmod/core/meta_data.py
@@ -169,7 +169,7 @@ class DataFormat(PydanticEnum):
     different partitioning schemes, making it useful to keep it as a thing unto itself.
     """
 
-    T_ROUTE_CONFIG = (13, {StandardDatasetIndex.HYDROFABRIC_ID: None}, None, False)
+    T_ROUTE_CONFIG = (13, {StandardDatasetIndex.DATA_ID: None, StandardDatasetIndex.HYDROFABRIC_ID: None}, None, False)
     """ Format for t-route application configuration. """
 
     @classmethod
--- a/python/lib/core/dmod/core/meta_data.py
+++ b/python/lib/core/dmod/core/meta_data.py
@@ -169,6 +169,9 @@ class DataFormat(PydanticEnum):
     different partitioning schemes, making it useful to keep it as a thing unto itself.
     """
 
+    T_ROUTE_CONFIG = (13, {StandardDatasetIndex.HYDROFABRIC_ID: None}, None, False)
+    """ Format for t-route application configuration. """
+
     @classmethod
     def can_format_fulfill(cls, needed: 'DataFormat', alternate: 'DataFormat') -> bool:
         """


robertbartel:
FYI @aaraney, I need the format value to exist for some of the composite dataset work I have going.  If it needs tweaking in any way to align with what you have been doing, just let me know.

aaraney:
I _think_ we need this too. Not to open too much of a can of worms, I am wondering as this enum grows if we should reconsider how we are modeling dataset formats in the future? [t-route's config file](https://github.com/NOAA-OWP/t-route/blob/master/doc/v3_doc.yaml) requires several input files depending on a simulations needs. Below are all the input files that _could_ be specified in a t-route config. Although, this may not be an exhaustive list, this is for `v3` of troute and they are now on `v4`.

- `preprocess_source_file`
- `geo_file_path`
- `mask_file_path`
- `level_pool_waterbody_parameter_file_path`
- `reservoir_parameter_file`
- `reservoir_rfc_forecasts_time_series_path`
- `lite_channel_restart_file`
- `lite_waterbody_restart_file`
- `wrf_hydro_channel_restart_file`
- `wrf_hydro_channel_ID_crosswalk_file`
- `wrf_hydro_waterbody_restart_file`
- `wrf_hydro_waterbody_ID_crosswalk_file`
- `wrf_hydro_waterbody_crosswalk_filter_file`
- `diffusive_domain`
- `topobathy_domain`
- `qlat_input_folder`
- `qlat_files`
- `usgs_timeslices_folder`
- `usace_timeslices_folder`
- `gage_segID_crosswalk_file`
- `wrf_hydro_lastobs_file`
- `gage_lakeID_crosswalk_file`
- `chanobs_filepath`
- `parity_check_input_folder`
- `validation_files`

For simplicity, are you thinking the `t-route` config file (the literal `yaml` file) and all the dependent input files would be part of a single composite dataset that _could_ also include realization configs for example?

robertbartel:
> For simplicity, are you thinking the t-route config file (the literal yaml file) and all the dependent input files would be part of a single composite dataset that could also include realization configs for example?

In part.  I need the notion of a t-route config dataset to exist, standing on its own.  The reason I need it is so that I can copy data items from it to a new composite configuration dataset.  Parts of the code for all that (and certainly parts beyond it eventually) require the _DataFormat_ value.  

> Not to open too much of a can of worms, I am wondering as this enum grows if we should reconsider how we are modeling dataset formats in the future?

Yeah, this is getting a bit tricky.  We should only need to care about things we consider indices of the format that go into defining its domain: i.e., the hydrofabric giving it context, the set of catchment ids for a realization config, etc.  So perhaps this needs a little more - like the index for the discrete list of files in the dataset - but I think we can still what we need out of this approach.

I'd like to continue treating configurations as data so we can store them in datasets and make them available the same way we do other data.  It is not easy to strictly define all the details of the config-related _DataFormats_ (like we would for a forcing format), but I am pretty sure it isn't necessary for how we use such things.  That said, if you have either a counter example or an alternative approach, we should consider it.

aaraney:
> In part. I need the notion of a t-route config dataset to exist, standing on its own. The reason I need it is so that I can copy data items from it to a new composite configuration dataset. Parts of the code for all that (and certainly parts beyond it eventually) require the _DataFormat_ value.

👍🏻

> Yeah, this is getting a bit tricky. We should only need to care about things we consider indices of the format...

Yeah, I agree with emphasis on the _we_. A listing of additional files seems good enough for now. It seems crucial that we have a good idea of the use cases _we_ need to support that rely on t-route in some way or else this area of the code is susceptible to scope creep.

> It is not easy to strictly define all the details of the config-related _DataFormats_ (like we would for a forcing format), but I am pretty sure it isn't necessary for how we use such things. That said, if you have either a counter example or an alternative approach, we should consider it.

Yeah, I agree for the most part. I hope we can ignore the majority of the details. The only counter example that comes to mind is deriving a t-route config for some simulation. It seems that we may not be able to select a `t-route` config on-the-fly, but instead it will require some human intervention to verify the knobs, dials, and input files are to the users requirements. I dont think that is an issue, so long as it falls within what we _need_ to support.

aaraney (approved):

PR #383, opened 2023-07-10
Addition of the InitialDataAdder abstraction

robertbartel:
A portion of a larger set of changes involved in work with moving to composite configuration dataset types.  Breaking this off parts like this that are fairly self contained for easier review.  

## Additions

- A new _InitialDataAdder_ abstract type and interface, used to add some initial data to a new dataset as it is being created

## Changes

- Modifying the _DatasetManager_ interface in its `create` function to use _InitialDataAdder_ instead of a _str_ for its `initial_data` param
- Modifying the _ObjectStoreManager_ implementation of _DatasetManager_ to have its `create` function comply with interface changes and use the `initial_data` object as a _InitialDataAdder_ appropriately.


aaraney:
At first glance, it appears that this is adding responsibility to dataset creation that creation should not be responsible for. Do you think could handle this use case if we instead introduced a dataset transaction or workflow abstraction? For starters we could limit the functionality to only operate on a single dataset. It _seems_ easier to model data transformation pipelines using a workflow (weaker guarantees than transaction) or transaction.

aaraney:
> My perspective is actually the opposite. This is creating an abstract type dedicated to obtaining and adding initial data to a dataset. Regarding whether dataset creation should be tied to adding data ...

Right, `create()` is not responsible for the logic of uploading initial data, but it is responsible for making and handling the call. It is responsible for handling its own failures and now any failures when adding initial data. To be fair, I think I am overthinking and over engineering this. In an ideal world, having an separate `Transaction` object that handles rollback logic akin to a db seems perfect for this problem. But, given our storage layer and the problem at hand, implementing a transaction abstraction is probably overkill.

aaraney (changes requested):
It looks like a minor change is needed to clean up a bucket's objects before removing the bucket in the case of a failure.

aaraney (approved):

PR #386, opened 2023-07-10
Upload data to dataset using readable object

aaraney:
Adds an async `upload_data_to_dataset` method to dataset clients that polls an async readable object that implements `async def read(self, size: Optional[int] = None) -> bytes`. This adds flexibility, so it is not required to treat to-be uploaded items as files.

## Additions

- Adds an async `upload_data_to_dataset` method to dataset clients that polls an async readable object that implements `async def read(self, size: Optional[int] = None) -> bytes`.

## Notes

- Not sure the correct _internal_ equivalent type to `MaaSDatasetManagementMessage`

## Todos

- unit tests
- implement `upload_data_to_dataset` for `DatasetInternalClient`

### DIFF
--- a/python/lib/client/dmod/client/_reader.py
+++ b/python/lib/client/dmod/client/_reader.py
@@ -1,4 +1,6 @@
-from typing import Protocol, Union
+import asyncio
+from functools import partial
+from typing import Optional, Protocol, Union
 
 
 class Reader(Protocol):
@@ -25,3 +27,17 @@ class AsyncReader(Protocol):
 
     async def read(self, size: Union[int, None] = ...) -> bytes:
         ...
+
+
+class AsyncReadWrapper:
+    """
+    Wrapper object for treating a sync reader as an async reader.
+    """
+
+    def __init__(self, reader: Reader) -> None:
+        self._reader = reader
+        self._loop = asyncio.get_running_loop()
+
+    async def read(self, size: Optional[int] = None) -> bytes:
+        fn = partial(self._reader.read, size)
+        return await self._loop.run_in_executor(None, fn)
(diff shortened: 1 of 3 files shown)


python/lib/client/dmod/client/_reader.py
```diff
@@ -0,0 +1,43 @@
...
+    ```
+    def upload(data: Reader) -> None:
+        ...
+
+    with open("some_file.txt", "rb") as fp:
+        upload(fp)
+    ```
+    """
+
+    def read(self, size: Union[int, None] = ...) -> bytes:
```
robertbartel, line 19: Any particular reason for `Union[int, None]` instead of `Optional[int]`?

robertbartel (changes requested):
I had a minor question and then a requested change.  We should be able to add an implementation for _DatasetInternalClient_ without too much trouble, so let's include it here.

python/lib/client/dmod/client/_reader.py
aaraney: Nothing in particular, I just took inspiration from python's BufferedReader interface:

```python
def read(self, __size: int | None = ...) -> bytes: ...
```

python/lib/client/dmod/client/request_clients.py
```diff
@@ -256,6 +280,10 @@ async def upload_to_dataset(self, dataset_name: str, paths: List[Path]) -> bool:
         # TODO: *********************************************
         raise NotImplementedError('Function upload_to_dataset not implemented')
 
+    async def upload_data_to_dataset(self, dataset_name: str, item_name: str, data: AsyncReader) -> bool:
+        # TODO
+        raise NotImplementedError('Function upload_data_to_dataset not implemented')
```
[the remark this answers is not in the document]
aaraney: Thanks for the guidance. Ill make that addition and let you know if I run into any roadblocks.

aaraney:
Closing this in favor of #540.

PR #391, opened 2023-07-14
Optionally disable DatasetExternalClient session caching and enable creation from existing session

aaraney:
## Additions

- Add an optional argument `cache_session` to the `DatasetExternalClient`'s initializer that makes it possible to opt out of cacheing session information on disk.
- Add `from_session` class method to create `DatasetExternalClient` from existing Session object.

### DIFF
--- a/python/lib/client/dmod/client/request_clients.py
+++ b/python/lib/client/dmod/client/request_clients.py
@@ -264,9 +264,11 @@ class DatasetExternalClient(DatasetClient,
     """
 
     # In particular needs - endpoint_uri: str, ssl_directory: Path
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, cache_session: bool = True, **kwargs):
         super().__init__(*args, **kwargs)
-        self._cached_session_file = Path.home().joinpath('.dmod_client_session')
+        self._cached_session_file: Optional[Path] = (
+            Path.home().joinpath(".dmod_client_session") if cache_session else None
+        )
 
     def _acquire_session_info(self, use_current_values: bool = True, force_new: bool = False):
         """
@@ -296,9 +298,16 @@ class DatasetExternalClient(DatasetClient,
             return tmp
 
     async def _async_acquire_session_info(self, use_current_values: bool = True, force_new: bool = False):
-        if use_current_values and not force_new and self._cached_session_file.exists():
+        if (
+            use_current_values
+            and not force_new
+            and self._cached_session_file is not None
+            and self._cached_session_file.exists()
+        ):
             try:
-                session_id, secret, created = self.parse_session_auth_text(self._cached_session_file.read_text())
+                session_id, secret, created = self.parse_session_auth_text(
+                    self._cached_session_file.read_text()
+                )
                 self._session_id = session_id
                 self._session_secret = secret
                 self._session_create = created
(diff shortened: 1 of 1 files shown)


python/lib/client/dmod/client/request_clients.py
```diff
@@ -264,9 +266,27 @@ class DatasetExternalClient(DatasetClient,
     """
 
     # In particular needs - endpoint_uri: str, ssl_directory: Path
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, cache_session: bool = True, **kwargs):
         super().__init__(*args, **kwargs)
-        self._cached_session_file = Path.home().joinpath('.dmod_client_session')
+        self._cached_session_file: Optional[Path] = (
+            Path.home().joinpath(".dmod_client_session") if cache_session else None
+        )
```
robertbartel, line 273: This wasn't the best `__init__` implementation.  It nullifies logic in the superclass for receiving a session file path, favoring a default value over a user-supplied one.  I don't remember if there was some reason other than laziness at this level.

Regardless, given there is logic in the _ExternalClient_ for __cached_session_file_, should we perhaps do this there?  Or, to phrase in a way that may be more scope-appropriate, why exactly does this type need to offer the option to not cache a session to a file.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -264,9 +266,27 @@ class DatasetExternalClient(DatasetClient,
...
+
+    @classmethod
+    def from_session(cls, *, endpoint_uri: str, ssl_directory: Path, session: Session, cache_session: bool = True, **kwargs) -> Self:
+        """
+        Create a `DatasetExternalClient` from an existing `Session` instance.
+
+        Note, the passed `Session` object will not be written to disk even if the `cache_session`
+        flag is present.  However, if the `Session` instance expires and a new session is acquired,
+        it will be cached if `cache_session` is set.
```
robertbartel, line 282: > Note, the passed `Session` object will not be written to disk even if the `cache_session` flag is present.

Took me a little to figure out why this was.  Seems to be another artifact of not reusing existing, inherited functionality very well.  I suspect we will need to address that.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -264,9 +266,27 @@ class DatasetExternalClient(DatasetClient,
...
+        """
```
robertbartel, line 283: Probably should include some docstring on the params here.

robertbartel (commented):
I have a larger, meta question for the changes that may need some discussion.  This probably will spawn several subsequent issues, though there's a small chance we'd need to do some of that here.   There's also one minor, non-essential docstring item.

python/lib/client/dmod/client/request_clients.py
aaraney: > Regardless, given there is logic in the _ExternalClient_ for __cached_session_file_, should we perhaps do this there? Or, to phrase in a way that may be more scope-appropriate, why exactly does this type need to offer the option to not cache a session to a file.

My intention was to use this client to create the rest proxy data service. In the context of that use case, an on disk session cache kind of gets in the way when all I really need to do is pass along an existing session's information.

My original plan was to remove the responsibility of caching and session management from the `_ExternalClient_` and introduce `AuthClient` and `CachedAuthClient` implementations. These objects would handle session logic like, for example, adding auth data to requests and in the future refreshing sessions. An external client could either have a `AuthClient` injected or create a default `AuthClient` object if one is not injected. However, this refactor was larger than I had anticipated and I am not confident that I could make theses changes without breaking existing behavior or introducing subtle bugs given the timeframe. So, I opted for the simplest solution.

robertbartel (approved):
This looks good.  We can look at doing a larger refactor later.

Issue #168, opened 2022-06-29
Update and clean up design for ngen worker image


robertbartel:
Two things jump out as needing to be addressed, though further things can also be done if appropriate.

First, the organization of the build stages, especially as they loosely align with the old _ngen-deps_ image (not used at the moment after moving away from Alpine), needs to be cleaned up.  There are also some redundancies that could be removed, and also probably some renaming of stages.

Second, the image no longer installs the ngen executable after building it.  Instead, it just sets things up to directly use the artifact from the CMake build directly.  This needs to be changed to installing the executable to reduce the image size.

robertbartel:
The specifics for this have evolved a little bit as the image has continued to be updated, but in general this type of cleanup is still needed.

PR #395, opened 2023-07-18
Move extra, test-specific layer of ngen worker Docker image

robertbartel:
Moving the _rocky_ngen_build_testing_ layer within the `ngen` image Dockerfile to be the next-to-last layer.  This will maximize what intermediate steps and layers can be tested with it, with minimal modifications to the layer's config, while minimizing what has to be rebuilt if/when the test layer is changed.

Loosely related to issue #168.

### DIFF
--- a/docker/main/ngen/Dockerfile
+++ b/docker/main/ngen/Dockerfile
@@ -413,14 +413,6 @@ RUN cd ${WORKDIR} \
 
 ################################################################################################################
 ################################################################################################################
-FROM rocky-ngen-deps as rocky_ngen_build_testing
-
-COPY --chown=${USER} --from=rocky_init_repo ${WORKDIR}/ngen ${WORKDIR}/ngen
-ENV BOOST_ROOT=${WORKDIR}/boost
-WORKDIR ${WORKDIR}/ngen
-
-################################################################################################################
-################################################################################################################
 ##### Create intermediate Docker build stage for building t-route in Rocky Linux environment
 FROM rocky-ngen-deps as rocky_build_troute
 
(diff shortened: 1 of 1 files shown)


docker/main/ngen/Dockerfile
```diff
@@ -575,7 +567,16 @@ RUN cd ${WORKDIR}/ngen \
                 $BUILD_DIR/test/test_bmi_multi; \
             fi \
     done \
-    && find cmake_build* -type f -name "*" ! \( -name "*.so" -o -name "ngen" -o -name "partitionGenerator" \) -exec rm {} + 
+    && find cmake_build* -type f -name "*" ! \( -name "*.so" -o -name "ngen" -o -name "partitionGenerator" \) -exec rm {} +
```
hellkite500, line 570: Hmm....so the intent is to build in all the test binaries in the previous layer and then just use them in the `testing` layer?  This will add some bloat to the ngen image size that may not be wanted long term.

robertbartel: I'm not entirely sure why this line registers in the diff.  Maybe I am missing it.  But regarding your question @hellkite500, none of the other layers should be built differently.  I'll be looking into those types of modifications, but that'll be handled separately.

This test layer is only used to build `ngen-build-test` image, defined as a separate service in _docker-build.yml_.  For the main image, this layer is ignored, because no other layer uses it, and it isn't the last layer.

The intent is just to move where in the build process this layer gets built.  That way, it's easier to control how the testing layer is used with minimal re-building of things that haven't really changed.

hellkite500 (approved):

Issue #396, opened 2023-07-18
Failing tests for dmod.evaluations package


robertbartel:
Several unit tests have begun to fail in _dmod.evaluations_.  This became apparent with the failure of checks in PR #394, although I am able to replicate the test failures locally using the `dev` branch. 

That said, I'm not quite sure what introduced this; checks and tests seem to have passed in the last closed PR, #385, and even in some other recent-but-not closed PRs, like #391.

robertbartel:
@christophertubbs, I'm initially assigning this to you, since you are most familiar with the package.  If you don't think you'll have time to get to this quickly, though, let me know.  I can pick it up if needed, and I don't want the failing checks to linger too long.

robertbartel:
Actually @christophertubbs, I've just noticed that the subsequent PR #395 didn't fail checks.  I'm going to do some more investigating myself, but feel free to pick this up if you have time. In particular, I'm curious if you can replicate failing tests locally on `dev` like I did.

christophertubbs:
What errors arose? I can't find a record of failing tests (probably just because I've never looked before).

robertbartel:
@christophertubbs, [here are the failed tests](https://github.com/NOAA-OWP/DMOD/actions/runs/5579729188/jobs/10217091914?pr=394#step:10:248) within the check of #394.

robertbartel:
FWIW, I re-ran `scripts/update_package.sh`, but this time added `-d` to also update all the project dependencies (i.e., stuff in _requirements.txt_).  After that, the evaluation tests are passing for me locally.

christophertubbs:
Mark me confused - I was just looking at the environment initialization text and it downloaded pandas 2.0.whatever, then pandas 1.5.whatever, then decided to build and install the wheel for 1.1.5? I'd venture to guess that an old version _would_ throw a wrench in there.

Would updating the root `requirements.txt` to read something like `pandas~=1.0` instead of just `pandas` do it? 2+ may be a little too recent and have breaking changes, but I imagine we'd want to keep up with the latest reasonable pandas version.

What's controlling that caching? I know it's the `_version.py` file for 90% of the subprojects, but what is it at the root?

christophertubbs:
`Timestamp.dayofyear` is just an alias for `Timestamp.day_of_year`, at this point, so nuking some underscores could do the trick.

robertbartel:
[...]

PR #397, opened 2023-07-18
Changed the `Timestamp.day_of_year` call to use the aliased `Timestamp.dayofyear` to work better with different versions of pandas

christophertubbs:
Older versions of pandas don't support `Timestamp.day_of_year` while new versions alias `Timestamp.dayofyear` to it. This will change the command used to make it work well with both old and new versions of pandas


christophertubbs:
Addresses #396

robertbartel (approved):

PR #400, opened 2023-07-20
Set definitions for what Django and Pandas versions to use

christophertubbs:
We ran into an issue the other day where imported dependencies were dictating that old versions should be used. I updated some of the version definitions to ensure that more modern versions of Pandas and Django are used.


robertbartel (commented):
I think in a vacuum, I'd prefer moving to Pandas 2.0, but I don't think we actually are compelled to at the moment (correct me if I'm wrong @christophertubbs).  On the other hand, we do have a transitive dependency on an earlier version of Pandas pending with #394.  

We could either leave this as blocked (until NOAA-OWP/ngen-cal#57 is addressed) or modify the Pandas version specified here to be something compatible with _ngen-cal_.  Thoughts?

robertbartel:
For now, I'm moving this to _Blocked_ while we wait on the aforementioned _ngen-cal_ issue.

aaraney:
@christophertubbs, @robertbartel, im blocking off time on Monday to investigate the `ngen-cal` issue (https://github.com/NOAA-OWP/ngen-cal/issues/57). Ill report back on this after im done with that.

aaraney:
Still waiting for https://github.com/NOAA-OWP/ngen-cal/pull/64 to be reviewed and then this should be unblocked.

aaraney:
https://github.com/NOAA-OWP/ngen-cal/pull/64 has been merged. Moving this to todo, now that this is unblocked.

PR #184, opened 2022-08-16
Evaluation service integration

christophertubbs:
Migration of evaluation service GUI functionality into the primary MaaS gui and the adoption of basic websocket handlers in the dmod.communication pattern.

PR #401, opened 2023-07-20
Standardize Evaluation ID generation

### LINKED ISSUE #215: Standardize evaluation id's
Just marking this so we can revisit in the future. The below comment should provide enough context.

_Originally posted by @christophertubbs in https://github.com/NOAA-OWP/DMOD/pull/184#discussion_r1014246624_

christophertubbs:
closes #215

Stems from [a comment in a PR](https://github.com/NOAA-OWP/DMOD/pull/184#discussion_r1014246624) about how example generated ids are created when rendering the editing screen for evaluations.

### DIFF
--- a/python/gui/MaaS/cbv/evaluation.py
+++ b/python/gui/MaaS/cbv/evaluation.py
@@ -34,9 +34,13 @@ class EvaluationListing(View):
         return render(request, template_name=self.template, context=context)
 
 
-def _generate_evaluation_id() -> str:
-    current_date = datetime.now()
-    date_representation = current_date.strftime("%m-%d_%H.%M")
+def _generate_evaluation_id(generation_date: datetime = None) -> str:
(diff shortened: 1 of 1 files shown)


christophertubbs:
The suggestions from the comments have been addressed.

christophertubbs:
The suggestions from the comments have been addressed. 

What needs to be further addressed is to what degree these ids need to be standardized and what all that means. Right now they are just identifiers so that User A can track what they ran and be able to describe its difference from what User B might be doing. The function in question just provides a default (currently just "manual_evaluation_at_{date}"). Using the current logic, "Evaluate HUC01" or "POD and FAR for Sugar Creek 10-19-2020" are just as valid.

aaraney (approved):
Looks good to me! Thanks, @christophertubbs!

PR #408, opened 2023-08-02
Cleanup and centralize ngen-related images

robertbartel:
Combining the builds for ngen, ngen-cal, and ngen partitioning into a single Dockerfile and build context.  Differentiating them through targeted Docker build stages.  Updating the scripts used by the images to combine parts of reusable logic, and to make other general improvements.

Depends on #405 and should not be merged until after that.


aaraney (approved):

PR #417, opened 2023-08-25
Transport Client Multiplexer

robertbartel:
Adding `TransportClientMultiplexer` type to wrap a `TransportLayerClient` instance in order to share the latter's connection, along with specialized `MuxStreamClient` concrete type that implements `TransportLayerClient` and is spawned from the wrapper to share the wrapped client.

This addresses the problem of users of a `TransportLayerClient` instance relying upon ordering of incoming messages in a way that is not guaranteed when there are multiple users. 

This PR relies on changes in #409 and should remain in draft status until that has been merged.


python/lib/communication/dmod/communication/client.py
```diff
@@ -504,6 +504,197 @@ def is_new_session(self) -> Optional[bool]:
...
+        ----------
+        data: Union[str, bytearray, bytes]
+            The data to send.
+        mux_id : str
+            A unique identifier for an isolated data stream.
+        """
+        if len(mux_id) != self._len_mux_id:
+            msg = f"{self.__class__.__name__} can't send using mux id {mux_id} (expected length {self._len_mux_id!s})"
+            raise ValueError(msg)
+        await self._wrapped_client.async_send(f"{mux_id}{data}")
```
aaraney, line 579: It seems that we will need to make changes to the request service's deserialization and serialization behavior to accommodate this change, right? One fear that I have is that this custom serialization format will spread across the entire codebase. For example, if you wanted to use this client and communicate with a particular service directly, that service would have to implement the custom deserialization format and handle stream ordering too.

aaraney (commented):
The implementation looks solid, @robertbartel. Unless I am missing something, it seems that the request service and likely all other services will need to be modified to accommodate handling stream ordering if we want to use a `TransportClientMultiplexer` to communicate with them, right? If that is the case, we should probably throw an exception when trying to create a `TransportClientMultiplexer` until those changes have been introduced. Thoughts?

robertbartel:
You are correct, @aaraney.  I got a little too focused on the looking glass and forgot about what was on the other side of it.

For now, I think we leave this as _Draft_/_Blocked_.  This is going to need a proper server-side solution to go along with it.  I'm not exactly sure what form that will need to take yet, and it may inform changes to what I initially had here.

aaraney:
> For now, I think we leave this as _Draft_/_Blocked_. This is going to need a proper server-side solution to go along with it. I'm not exactly sure what form that will need to take yet, and it may inform changes to what I initially had here.

Okay, cool! Yeah, im not sure either. Ill spend some time thinking about what that might look like too and maybe we can come together and compare our thoughts.

PR #420, opened 2023-09-07
Fix support for dependency version clauses for certain Python dependencies in Docker images 

robertbartel:
We have to be specific regarding certain Python dependencies - pandas for example.  Additionally, a subset of such packages also each take a long time to install within our Docker images, leading us to have set up the images/stages to build wheels for these packages in advance and in isolated build stages.  

There were Dockerfile `ARG` elements set up to control the version clause of these packages if desired, but nothing in place to actually supply values to those.  These changes address that by updating the stack compose config file to supply those `ARG`s from the _.env_ config.

Also added is another _.env_-controlled flag regarding whether the aforementioned version clause `ARG`s should override an explicit version clause within the project-level _requirements.txt_ file, which is also used by the builds.

### DIFF
--- a/docker/py-sources/docker-build.yml
+++ b/docker/py-sources/docker-build.yml
@@ -6,6 +6,14 @@ services:
     build:
       context: ../../
       dockerfile: ./docker/py-sources/py-deps.Dockerfile
+      args:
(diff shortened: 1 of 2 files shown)


aaraney (approved):
Looks good to me! The PR documentation made this really easy to review. Thanks as always, @robertbartel. Merge at will!

PR #421, opened 2023-09-07
Django Testing Additions

christophertubbs:
Added specialized testing functionality for all django applications


scripts/test_django.py
```diff
@@ -0,0 +1,702 @@
...
+"""
+The root of this application. Path(__file__) will be ./DMOD/scripts/test_django.py, 
+Path(__file__).parent will be ./DMOD/scripts, so you have to hit Path(__file__).parent.parent 
+to hit the root, i.e. ./DMOD
+"""
+
+DEFAULT_ROOT = str(APPLICATION_ROOT / "python")
+"""The root of the python libraries and services"""
+
+MANAGE_MARKER = "os.environ.setdefault('DJANGO_SETTINGS_MODULE'"
```
aaraney, line 27: Is there a missing closing paren here?

aaraney:
```suggestion
MANAGE_MARKER = "os.environ.setdefault('DJANGO_SETTINGS_MODULE')"
```

scripts/test_django.py
```diff
@@ -0,0 +1,702 @@
...
+            quiet: Whether to print in quiet mode
+        """
+        # verbose and quiet mode conflict, so throw an error if they are encountered
+        if verbose and quiet:
+            print(
+                "Output cannot be both quiet and verbose; choose either '--quiet' or '--verbose', but not both",
+                file=sys.stderr
+            )
+            # Exit with a code of -1 to indicate that this was an application error, not a test error or failure
+            exit(-1)
```
aaraney, line 370: `POSIX` exit codes are unsigned 8 bit integers. Fortunately, python is smart enough and wraps this around to `255`, but you may want to change this to some other meaningful positive number.

scripts/test_django.py
```diff
@@ -0,0 +1,702 @@
...
+                    current_content += os.linesep if line == LAST_END_MESSAGE else f"{os.linesep}{line}"
+
+        # If it's detected that we're still parsing and have reached the end of the document, we want to attach 
+        # what we have as a new message
+        if current_content and in_message:
+            current_content = current_content.strip()
+            message = TestMessage(status=current_status, content=current_content, description=current_description)
+            self.messages.append(message)
+
+    def print(self, verbose: bool = None, quiet: bool = None):
```
aaraney, line 355: I think both `verbose` and `quiet`'s default value should be `False` instead of `None` here.

aaraney (changes requested):
I have one minor request and then we should be on the way. Thanks, @christophertubbs!

scripts/test_django.py
christophertubbs: That won't work - the marker is supposed to match on lines like `x = os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'something.settings')`. Cutting it off with the `)` character will prevent it from matching on what it needs to match on.

aaraney (approved):

aaraney:
Thanks @christophertubbs! Merge at will!

aaraney (dismissed):

aaraney (approved):

PR #424, opened 2023-09-08
Improve dmod.client and update CLI

robertbartel:
General improvements to the CLI, as well as updates to bring in line with more recent DMOD updates to serialization, communication client classes, and user config representation.

Note that this PR depends on #409 and should remain in draft status until that is merged.


python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,600 +1,823 @@
...
         with dest.open('w') as file:
-            for page in range(1, (self.last_response.total_pages + 1)):
-                request = DatasetManagementMessage(action=ManagementAction.DOWNLOAD_DATA, dataset_name=dataset_name,
-                                                   data_location=item_name, page=page)
-                self.last_response: DatasetManagementResponse = await self.async_make_request(request)
-                file.write(self.last_response.file_data)
+            # Do initial request outside of generator
+            await self._transport_client.async_send(data=str(request))
+            async for received_data_msg in self._transfer_receiver():
+                file.write(received_data_msg.data)
```
aaraney, line 348: It is guaranteed the full buffer passed into `write()` is flushed to the file. I think we need to loop and call write repeatedly until the full buffer has been written.

```suggestion
            async for received_data_msg in self._transfer_receiver():
                data = received_data_msg.data
                while data:
                    bytes_written = file.write(data)
                    data = data[bytes_written:]
```

python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,600 +1,823 @@
...
+
+        try:
+            request, final_response_type = self._request_prep(dataset_name=dataset_name, item_name=item_name,
+                                                              action=ManagementAction.REQUEST_DATA)
+        # TODO: (later) implement and use DmodAuthenticationFailure
+        except DmodRuntimeError as e:
+            reason = f'{self.__class__.__name__} Download Auth Failure'
+            return MaaSDatasetManagementResponse(success=False, reason=reason, message=str(e))
+
         with dest.open('w') as file:
```
aaraney, line 344: I think we need to write bytes here (if we are going to handle opening the file)?

python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,600 +1,823 @@
...
+            msg = f"{self.__class__.__name__} failed with {e.__class__.__name__} parsing `{final_data}` to JSON)"
+            return final_response_type(success=False, reason=f"JSON Parse Failure On Final Response", message=msg)
+
+        final_response = final_response_type.factory_init_from_deserialized_json(final_response_json)
+        if final_response is None:
+            return final_response_type(success=False, reason="Failed to Deserialize Final Response")
+        else:
+            return final_response
+
+    async def upload_dataset_item(self, dataset_name: str, item_name: str, source: Path) -> DatasetManagementResponse:
```
aaraney, line 363: Akin to my other comments, we should consider using a `Reader` as a source instead of a `Path` object.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,600 +1,823 @@
...
+                                                              action=ManagementAction.ADD_DATA)
+        # TODO: (later) implement and use DmodAuthenticationFailure
+        except DmodRuntimeError as e:
+            reason = f'{self.__class__.__name__} Upload Request Auth Failure'
+            return MaaSDatasetManagementResponse(success=False, reason=reason, message=str(e))
 
-    async def upload_to_dataset(self, dataset_name: str, paths: List[Path]) -> bool:
+        chunk_size = 1024
+
+        with source.open('r') as file:
```
aaraney, line 377: I think we need to open this in reading bytes, right? Since we dont know what kind of item is being uploaded.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,600 +1,823 @@
...
+            last_send = False
+            raw_chunk = file.read(chunk_size)
+
+            while True:
+                await self._transport_client.async_send(data=str(message))
```
aaraney, line 382: I didnt see a `message` variable in scope. I think you meant, right?

```suggestion
                await self._transport_client.async_send(data=raw_chunk)
```

aaraney: Ahh, I see I think the `message` variable needs to be moved up, right?

```suggestion
            message = DataTransmitMessage(data=raw_chunk, is_last=last_send)
            while True:
                await self._transport_client.async_send(data=str(message))
```

python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,600 +1,823 @@
...
-    def _process_data_download_iteration(self, raw_received_data: str) -> Tuple[bool, Union[DataTransmitMessage, MaaSDatasetManagementResponse]]:
+        See Also
+        -------
+        upload_to_dataset
+        uses_auth
         """
-        Helper function for processing a single iteration of the process of downloading data.
+        request = DatasetManagementMessage(action=ManagementAction.CREATE, domain=domain, dataset_name=name,
+                                           category=category)
+        create_response = await self._process_request(request=request, originating_func_name='create_dataset')
```
aaraney, line 556: Just a general comment. I get the purpose of `originating_func_name`, but it is raising my code smell alarms.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,600 +1,823 @@
...
+        list_datasets
         """
-        # TODO: think about if anything is needed for this
-        pass
+        action = ManagementAction.LIST_ALL if category is None else ManagementAction.SEARCH
+        request = DatasetManagementMessage(action=action, category=category)
+        return await self._process_request(request=request, originating_func_name='get_dataset_names')
 
-    async def _upload_file(self, dataset_name: str, path: Path, item_name: str) -> bool:
+    async def get_dataset_items(self, dataset_name: str, **kwargs) -> DatasetManagementResponse:
```
aaraney, line 637: Maybe `get_dataset_items_info`? get items makes me think I am downloading things.

python/lib/client/dmod/test/test_dataset_client.py
```diff
@@ -1,46 +1,48 @@
...
-class SimpleMockDatasetClient(DatasetClient):
+class SimpleMockDataServiceClient(DataServiceClient):
     """
-    Mock subtype, primarily for testing base implementation of ::method:`_parse_list_of_dataset_names_from_response`.
+    Mock subtype, primarily for testing base implementation of ::method:`extract_dataset_names`.
     """
 
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+        #super().__init__(*args, **kwargs)
```
aaraney, line 14: Minor: could be removed.

aaraney (changes requested):
Thanks @robertbartel! I have a few fairly minor comments that more or less boil down to API parameter choices. There is one bug that will need to be addressed but other than that, this should be pretty easy to get through. Thanks!

python/lib/client/dmod/client/dmod_client.py
```diff
@@ -1,277 +1,213 @@
...
 
-        In the event neither a domain not a data format is provided, a ::class:`ValueError` is raised.
+    @property
+    def client_config(self) -> ClientConfig:
+        return self._client_config
 
-        Additionally, keyword arguments are forwarded in the call to the ::attribute:`dataset_client` property's
-        ::method:`DatasetClient.create_dataset` function.  This includes the aforementioned kwargs for a creating a
-        default ::class:`DataDomain`, but only if they are otherwise ignored because a valid domain arg was provided.
+    async def data_service_action(self, action: str, **kwargs) -> ResultIndicator:
```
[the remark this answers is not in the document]
robertbartel: I'm not sure it will be as valuable as it might first appear, because of the way the class is used, but it's trivial enough that I'm ok with adding it.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,600 +1,823 @@
...
-    def _parse_list_of_dataset_names_from_response(self, response: DatasetManagementResponse) -> List[str]:
+        Returns
+        -------
+        ResultIndicator
+            An indicator of success of the request that, when successful, contains he full state of the provided job,
+            formatted as a JSON dictionary, in the ``data`` attribute.
         """
-        Parse an included list of dataset names from a received management response.
+        # TODO: implement
+        raise NotImplementedError('{} function "request_job_info" not implemented yet'.format(self.__class__.__name__))
```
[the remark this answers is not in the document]
robertbartel: Sure: see #431.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,600 +1,823 @@
...
+            An indicator that, when successful, includes as ``data`` the list of ids of existing jobs.
+
+        See Also
+        -------
+        get_jobs_list
+        """
+        # TODO: implement
+        raise NotImplementedError('{} function "request_jobs_list" not implemented yet'.format(self.__class__.__name__))
+
+    async def submit_ngen_request(self, **kwargs) -> NGENRequestResponse:
```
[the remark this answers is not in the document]
robertbartel: @aaraney, the main reason was that - thinking from the perspective of CLI usage - I didn't want _DmodClient_ to have to be responsible for message initialization or deserialization.  The specialized client needed to handle deserialization.  Given the usage setup, this required it to be capable of just accepting passed-through CLI parsed args (and required `__main__` to be somewhat particular about the names of the parsed args it allowed).

That said, it does seem as though either the functions are misnamed or they need to _also_ accommodate complete objects.  Would it be too messy if, say, I checked for the presence of a `request` key in `kwargs` and used that directly if it was request object?

python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,600 +1,823 @@
...
+        Parameters
+        ----------
+        job_type : str
+            String representation of the type of request: either ``ngen`` or ``ngen_cal``.
+        request_file : Path
+            The supplied file containing a JSON string, which should be a serialized, applicable request object.
+        """
+        return await self.submit_request_from_json(job_type=job_type, request_json=request_file.read_text())
+
+    async def submit_request_from_json(self, job_type: str, request_json: Union[dict, str], **kwargs) -> ResultIndicator:
```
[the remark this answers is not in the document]
robertbartel: Need to do this slightly differently, but sure.

python/lib/client/dmod/client/request_clients.py
```diff
@@ -1,600 +1,823 @@
...
+            return NgenCalibrationResponse.factory_init_from_deserialized_json(
+                json.loads(await self._submit_job_request(NgenCalibrationRequest.factory_init_from_deserialized_json(request_json))))
+        else:
+            raise RuntimeError(f"Invalid job type indicator for serialized job request: {job_type}")
+
+
+class DataTransferAgent(ABC):
+
+    @abstractmethod
+    async def download_dataset_item(self, dataset_name: str, item_name: str, dest: Path):
```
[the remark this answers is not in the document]
robertbartel: Indeed, I had a similar idea in mind, and even started a bit of work for it, but I decided it would be too much trouble for the time being to finish it correctly.  I have created #432 to track the task in the future.

python/lib/client/dmod/client/request_clients.py
robertbartel: Hmm.  I'll have to spend some time reading through the `io` module docs, but I thought that use of the context manager protected against this.  I.e., that `bytes_written` would always equal `len(data)` on the first iteration of your suggested `while` loop, and that the potential risk was whether the OS actually _wrote_ things to disk (which use of the context manager ensures).  

That said, I am ok with adding this to be on the safe side.

python/lib/client/dmod/client/request_clients.py
robertbartel: Not sure I follow.  By the time data to write gets here, it's a `str` attribute of an inflated object.  Are you suggesting it would be a better practice to deal with files in binary mode and just cast these `str`s back to `bytes`?  Or perhaps a larger reworking that keeps the data as `bytes` somehow?

python/lib/client/dmod/client/request_clients.py
robertbartel: Hmm ... yeah, I believe you are correct, and as such we should do similarly for writes as you suggested above.   But this actually becomes a larger undertaking.   All the links in the chain will need to be modified to accommodate, from _DataTransmitMessage.data_ to the declaration and implementations of _DatasetManager.add_data()` (and associated helper methods).  

It probably makes sense to do this as part of #432.  What do you think?

python/lib/client/dmod/client/request_clients.py
robertbartel: Good catch, although your solution isn't quite right.  The initial _DatasetManagementMessage_ is created by the earlier call to `_request_prep`.  Because we need to look ahead in the data in order to figure out whether we are doing our "last_send," both the initial request and the initial "current" chunk of data are prepared outside the loop, and the subsequent message is prepared at the end of the loop.

I'll fix momentarily.

python/lib/client/dmod/client/request_clients.py
robertbartel: Indeed, it was less that ideal. I've refactored things.

python/lib/client/dmod/client/request_clients.py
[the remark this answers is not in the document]
robertbartel: Pinging on this again.

python/lib/client/dmod/client/request_clients.py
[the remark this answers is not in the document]
robertbartel: Yeah, as above, looks like this approach would soon be deprecated.

python/lib/client/dmod/client/request_clients.py
robertbartel: @aaraney, pinging on this follow up question again.

python/lib/client/dmod/client/request_clients.py
robertbartel: This should fall under #432 again.

python/lib/client/dmod/client/request_clients.py
robertbartel: Ping again here also.  This probably relates to the earlier item and my old follow-up questions.

python/lib/client/dmod/client/request_clients.py
robertbartel: Just to confirm, because the variable in question is initialized prior to these specific lines, this comment doesn't get attributed as _Outdated_, but I have applied a fix.

python/lib/client/dmod/client/dmod_client.py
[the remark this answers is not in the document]
aaraney: Thanks for doing some digging. Determining what is supported in x version of python is always tedious. The general community guidance so far is to use [`typing_extensions`](https://github.com/python/typing_extensions). This is a library that ports `python.typing` features found in later version of python. We are already using it in the client, communication, core, and scheduler libraries to extend support. I would recommend that we use it in this case too.

The python 3.12 question is a much larger one and IMO a can of worms. `3.12` introduces a new `type` keyword that lets you effectively create what are now `TypeAlias`. However, until `3.12` is our minimum version we will need either `typing.TypeAlias` or `typing_extensions.TypeAlias`. So, to me, the 3.12 question it is worth punting on for now. Thoughts?

python/lib/client/dmod/client/request_clients.py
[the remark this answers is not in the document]
aaraney: > the main reason was that - thinking from the perspective of CLI usage - I didn't want _DmodClient_ to have to be responsible for message initialization or deserialization.

Yeah I hear that argument and I agree. I don't think either `DmodClient` nor the `JobClient` should be responsible for message initialization or deserialization. After taking a step back, I think this boils down to a difference in option of when to use factory functions. `DmodClient.execute_job` is a great example of this. Below is the signature for convenience sake:

```python
    async def execute_job(self, workflow: str, **kwargs) -> ResultIndicator:
```

IMO I would have implemented it as something like:

```
    async def execute_job(self, job: ExternalAbstractNgenRequest) -> ResultIndicator:
```

Right the burden of initialization or deserialization is now put on whoever uses a `DmodClient`, but IMO that is a good thing. If the code that _uses_ a `DmodClient` needs to do some custom logic to transform a request or deserialize the request from different sources, that is up to the code that _uses_ the client. It also simplifies how code that _uses_ a `DmodClient` handle failures. With the current implementation an `execute_job` can fail for numerous reasons. try catching would be a little painful to disseminate good error messages back to a user. I think we can simplify the `DmodClient` and `JobClient` by shifting the burden of initialization / deserialization to the code that _uses_ a `DmodClient`. 

You know me, I am a big fan of dependency injection and sometimes overly so. You have a lot more experience developing software than I do and I really value you thoughts. So I dont mean to drag this on, but it would be really valuable for me to hear your argument for using a factory here.

Im happy to keep it as is, but I do hope to learn something from our discussion!

python/lib/client/dmod/client/request_clients.py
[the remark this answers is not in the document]
aaraney: Awesome, thanks for doing that!

python/lib/client/dmod/client/request_clients.py
aaraney: > Or perhaps a larger reworking that keeps the data as bytes somehow?

I didn't realize that `DataTransmitMessage`'s `data` field is `str`. Yeah I think we need to change the `data` field to either `Union[str, bytes]` or just `bytes`. Without accepting data as `bytes` we will get `UnicodeDecodeError` exceptions for any non-unicode data. 

For writing to the file, we could either cast everything we get to `bytes` or do a type check if it is `str` or `bytes` and open the file in the correct mode accordingly.

python/lib/client/dmod/client/request_clients.py
aaraney: Yeah I agree. This is a larger undertaking and not directly related so should be in its own PR. 

I opened #486 to track changing `DataTransmitMessage`'s `data` field to a more appropriate type. Im not sure if there are other models off the top of my head that will also need to be changed.

python/lib/client/dmod/client/request_clients.py
aaraney: Awesome, yeah you fixed what I was talking about. Thanks!

python/lib/client/dmod/client/request_clients.py
aaraney: Thanks!

aaraney:
@robertbartel, thanks for revitalizing this! I'd forgotten about it and let it get stale. I've resolved relevant comments and contributed to existing conversation to move this forward. We are getting really close, just a few minor things.

python/lib/client/dmod/client/request_clients.py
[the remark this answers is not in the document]
robertbartel: > I don't think either DmodClient nor the JobClient should be responsible for message initialization or deserialization. .... Right the burden of initialization or deserialization is now put on whoever uses a DmodClient, but IMO that is a good thing.

I'm not sure that makes sense.  I think of _DmodClient_ as intended for use by something outside of DMOD (for purposes of this discussion, I think it reasonable to treat the functions handling interfacing with the command line as on the "outside").   We could still put the (application logic) burden on them, but then what purpose do these client classes serve?  And would it not be better to provide something that encapsulates this behavior (if so, does that belong somewhere else)?

python/lib/client/dmod/client/request_clients.py
robertbartel: It didn't occur to me back in September, but (unless I'm missing something) we can't actually do this because it will break JSON serialization.

python/lib/client/dmod/client/request_clients.py
robertbartel: As mentioned elsewhere, `data` is a `str` in order to support serialization of the messages.

python/lib/client/dmod/client/request_clients.py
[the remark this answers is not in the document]
aaraney: > We could still put the (application logic) burden on them, but then what purpose do these client classes serve? And would it not be better to provide something that encapsulates this behavior (if so, does that belong somewhere else)?

I see your point. I've been thinking of these client classes more like an RPC client where the main purposes of the client are to handle connection, session state, transmission, and reception of data. Not like an HTTP request client where the underlying protocol is flat in nature and easier to abstract. In either case, with the existing API there is nothing stopping you from creating a request object (pydantic model), dumping that to a dictionary, and passing that in as input. I assumed building up a request would be the primary way to interface with the api (or deserializing user input and then passing in the object).

The one issue with passing in a request object is mutating / attaching session information. We've run into this in other contexts before. I am not sure if accepting `**kwargs` makes that any easier or has any benefits.

python/lib/client/dmod/client/request_clients.py
aaraney: Yeahhhhh, you are right. I forgot about that. Does that mean you can only upload and download unicode encoded data at this moment? I don't recall seeing any logic to base64 / base85 encode binary data.

python/lib/client/dmod/client/request_clients.py
aaraney: Not to open a can of worms, but as mentioned elsewhere, does this effectively limit us to uploading and downloading unicode encoded data?

python/lib/client/dmod/client/request_clients.py
robertbartel: Just as a sanity check, if we decode and then encode binary data using the "wrong" encoding, but it's the same encoding on both ends, won't we get the same data we started with?

Assuming more discussion and work is needed, I suggest that we defer this problem.  A solution will probably be complex, and we can find workarounds as needed to get datasets into the system (worst case: temporarily opening up MinIO externally and uploading directly).

python/lib/client/dmod/client/request_clients.py
robertbartel: I modified #486 a bit to involve a large process related to this, though as I mentioned elsewhere, I think we kick this can down the road for now.  We can work around it if necessary, and we need to prioritize other tasks.

python/lib/client/dmod/client/request_clients.py
aaraney: > Just as a sanity check, if we decode and then encode binary data using the "wrong" encoding, but it's the same encoding on both ends, won't we get the same data we started with?

Yeah! You are right. I may be blowing this out of proportion and it might not be a problem after all. I thought to transmit _any_ data to and from the request service (and therefore the data service) it must be in json and therefore in some unicode flavor (e.g. utf-8). So, for example if I wanted to upload a netCDF forcing file to a dataset, _I think_ I would first need to encode and chunk the data using base64 or base85, package that up in json, and send it along, right? I agree that we should move this discussion elsewhere and just focus on getting the PR through lol. Sorry to trying to derail us 😅!

python/lib/client/dmod/client/dmod_client.py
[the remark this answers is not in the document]
robertbartel: Let's leave this out for now.  It's non-essential, and it adds eventual technical debt with an obfuscated timeline.  On top of that, it doesn't cover the entire scope of where these strings are relevant; i.e., it may leave a trap to fall into later when CLI arg parsing needs some changes.

We can still consider it, but let's look at that separately.

aaraney (approved):

robertbartel:
The only thing different was manually resolving a conflict for a version bump to the communication package, and the result of that ended up the same.  As such, I'm going to bypass waiting for another approval this time.

Issue #425, opened 2023-09-11
Correct issues with build of Docker images for DMOD services and Python sources


robertbartel:
At present, Docker images for the `py-sources` stack do not build successfully.  These build and assemble dependencies for the DMOD service Docker images in the `main` stack (e.g., that of the scheduler service).  The result is that new service images also cannot be built.

Initially, there are at least two things causing trouble (though there could be others not yet fully apparent): 
- recent changes in later versions of `pip` and in how Python package are built (as discussed in [this ngen issue](https://github.com/NOAA-OWP/ngen/pull/516#issuecomment-1523692130))
- usage of a rather old base image: `python:3.8-alpine3.15`

Note also that it is anticipated that there will be trouble with service images related to the image builds not completely accounting for version restrictions of certain dependencies (e.g. _pandas_) as required by some of the DMOD internal packages (see #400), though strictly speaking, we haven't had that trouble yet.

robertbartel:
PR #420 introduces a fix that it was hoped would resolve the image build problems this issue tracks.  It did not, though the changes are still worth doing and could still be a necessary part of the entire eventual solution.

PR #427, opened 2023-09-12
Move DMOD service images and service Python packaging image to Rocky Linux base 

robertbartel:
Move `main` stack service images and the `py-sources` image (on which the service images depend) to a base of Rocky Linux.  This removes the need for several pre-built wheels and simplifies the build overall, at the cost of slightly larger images.  

These changes resolve some issues causing image builds to currently fail, as tracked in #425.


aaraney:
Is this ready to review or are you considering targeting rocky 9 after our conversation earlier, @robertbartel? For a little context, there is an ongoing discussion if NextGen needs to support GCC 8. This discussion centers around the absence of c++ std smart pointer types in GCC 8. Coincidentally, rocky linux 8's package repo ships GCC 8.5 where as rocky linux 9 ships GCC 11.

robertbartel:
I am still considering moving the _ngen_ images (and the others) to 9.1, but it turned out that 9.1 didn't help with the problem I was having. Since it isn't immediately helpful/necessary, I don't want to rush to make that change.  

For now, let's move forward with 8.5 on this one.

aaraney (approved):
This looks good to me! Thanks, @robertbartel!

PR #318, opened 2023-04-10
Change routing build to t-route master branch

mattw-nws:
Switch to building/installing t-route master instead of t-route ngen branch.

## Additions

- Several small build system improvements, including using `${BUILD_PARALLEL_JOBS}` instead of `$(nproc)`, and a way to optionally pass a pre-downloaded boost tarball to the image creation process.

## Removals

- Does not work out of the box to use legacy t-route (e.g. ngen branch), which might limit capability with some older hydrofabric datasets.

## Changes

- Now compiles and builds master t-route by default (instead of ngen branch)
- Several Python elements are now installed in the global (root) site. This enables use-cases where a user other than mpi(1000) is used, e.g. using `docker run --user=$(uid -u)`, so now the routing modules are available to that user (tested!).

## Testing

1. Built and ran the `gauge_01073000` dataset, including using `--user` to change the uid

## Notes

-

## Todos

-

## Checklist

- [X] PR has an informative and human-readable title
- [X] Changes are limited to a single goal (no scope creep)
- [X] Code can be automatically merged (no conflicts)
- [X] Code follows project standards (link if applicable)
- [X] Passes all existing automated tests
- [X] Any _change_ in functionality is tested
- [X] New functions are documented (with a description, list of inputs, and expected output)
- [X] Placeholder code is flagged / future todos are captured in comments
- [X] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [X] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [X] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Target Environment support

- [X] Linux

PR #429, opened 2023-09-13
Update ngen Docker image to build t-route master branch

robertbartel:
Switch to building/installing t-route master instead of t-route ngen branch.  Very similar to changes from #318, but with some modifications based on initial review and evolution of the repo since.

## Additions

- Several small build system improvements including: 
  - using `${BUILD_PARALLEL_JOBS}` instead of `$(nproc)`
  - an improved way to optionally pass a pre-downloaded boost tarball to the image creation process, without reducing Docker build cache utilization
  - remove of the requirement for t-route repo/branch values to be explicitly provided in the .env file

## Removals

- Does not work out of the box to use legacy t-route (e.g. ngen branch), which might limit capability with some older hydrofabric datasets.

## Changes

- Now compiles and builds master t-route by default (instead of ngen branch)
- Several Python elements are now installed in the global (root) site. This enables use-cases where a user other than mpi(1000) is used, e.g. using docker run --user=$(uid -u), so now the routing modules are available to that user
  - These changes were carried over from #318; they were tested there but have not been subsequently retest as of this writing

### DIFF
--- a/docker/main/docker-build.yml
+++ b/docker/main/docker-build.yml
@@ -70,8 +70,8 @@ services:
         REPO_URL: ${NGEN_REPO_URL?No NGen repo url configured}
         BRANCH: ${NGEN_BRANCH?No NGen branch configured}
         COMMIT: ${NGEN_COMMIT}
-        TROUTE_REPO_URL: ${TROUTE_REPO_URL?No t-route repo url configured}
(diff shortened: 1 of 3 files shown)


aaraney (approved):
Looks good to me!

PR #430, opened 2023-09-18
Fix and optimize ngen-related image builds

robertbartel:
Address a few problems and add some optimizations to the ngen worker, calibration worker, partitioner Docker images:

- Move to RockyLinux 9.1 to get default support for GCC 11
  - Needed in particular to work around some current issues in the framework
- Update version of mpich to play nicely with Fortran from GCC 11
- Correct Boost dependency version to be recent enough for current framework requirements (now 1.82.0)
- Modify build stages that build ngen to (typically) do a `git pull` just before to ensure working on most recent commit for a branch
- Ensuring Docker stack config, which controls the overall build process, establishes the same set of `ARG` values for related images (`ngen`, `ngen-calibration`, etc.) so that things behave as expected/consistently and the build cache is used effectively
- Ensure use of `pip`'s `--no-cache-dir` option to optimize storage space a bit

### DIFF
--- a/docker/main/docker-build.yml
+++ b/docker/main/docker-build.yml
@@ -59,7 +59,17 @@ services:
       context: ./ngen
       target: rocky_ngen_build_testing
       args:
+        REPO_URL: ${NGEN_REPO_URL?No NGen repo url configured}
+        BRANCH: ${NGEN_BRANCH?No NGen branch configured}
+        COMMIT: ${NGEN_COMMIT}
+        TROUTE_REPO_URL: ${TROUTE_REPO_URL:-https://github.com/NOAA-OWP/t-route}
+        TROUTE_BRANCH: ${TROUTE_BRANCH:-master}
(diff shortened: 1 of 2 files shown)


docker/main/docker-build.yml
```diff
@@ -59,7 +59,17 @@ services:
       context: ./ngen
       target: rocky_ngen_build_testing
       args:
+        REPO_URL: ${NGEN_REPO_URL?No NGen repo url configured}
```
aaraney, line 62: What are your thoughts on being more explicit here and naming these `NGEN_REPO_URL`, `NGEN_BRANCH`, and `NGEN_COMMIT` instead? If I were looking at a `.env` file that contained this information and saw `REPO_URL` I would be a little confused about which repo to specify.

aaraney (approved):
I have one question that should be addressed, but I dont think it necessarily needs to hold this up. Thanks, @robertbartel!

PR #433, opened 2023-09-19
Fix request service handling of eval service

robertbartel:
Deals with some issues in how the request service deals with evaluation service connection details, as the latter doesn't actually exist in the `main` deployment stack yet.  These are currently keeping the request service from starting successfully.

- Fix request service main package arg handling, which was overwriting args related to data service connections with eval service values
- Within _RequestService_, wrap creation of evaluation request handler in `try` block, falling back to a `None` value if something is raised 
- Raise a _RuntimeError_ in _RequestService.listener()_ if an evaluation request comes in but the evaluation request handler attribute is `None`.

### DIFF
--- a/python/services/requestservice/dmod/requestservice/service.py
+++ b/python/services/requestservice/dmod/requestservice/service.py
@@ -144,13 +144,18 @@ class RequestService(WebSocketSessionsInterface):
                                                            service_host=data_service_host,
                                                            service_port=int(data_service_port),
                                                            service_ssl_dir=self.data_service_ssl_dir)
-
-        self._evaluation_service_handler = EvaluationRequestHandler(
-            target_service='evaluation-service',
-            service_host=evaluation_service_host,
-            service_port=evaluation_service_port,
-            ssl_directory=evaluation_service_ssl_dir
-        )
+        # This probably won't work until evaluation service is properly added in Docker stack, so wrap in try for now
+        try:
+            self._evaluation_service_handler = EvaluationRequestHandler(
+                target_service='evaluation-service',
+                service_host=evaluation_service_host,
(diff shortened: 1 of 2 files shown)


python/services/requestservice/dmod/requestservice/service.py
```diff
@@ -170,6 +175,10 @@ async def listener(self, websocket: WebSocketServerProtocol, path):
                 event_type = MessageEventType.INVALID if req_message is None else req_message.get_message_event_type()
 
                 if isinstance(req_message, LaunchEvaluationMessage) or isinstance(req_message, OpenEvaluationMessage):
```
christophertubbs, line 177: `if isinstance(req_message, (LaunchEvaluationMessage, OpenEvaluationMessage))` instead of `if isinstance(req_message, LaunchEvaluationMessage) or isinstance(req_message, OpenEvaluationMessage)` might read better here.

python/services/requestservice/dmod/requestservice/service.py
```diff
@@ -144,13 +144,18 @@ def __init__(self, listen_host='',
...
+        try:
+            self._evaluation_service_handler = EvaluationRequestHandler(
+                target_service='evaluation-service',
+                service_host=evaluation_service_host,
+                service_port=evaluation_service_port,
+                ssl_directory=evaluation_service_ssl_dir
+            )
+            self._eval_handler_exception = None
+        except Exception as e:
+            self._evaluation_service_handler = None
```
christophertubbs, line 157: Might be worth adding some logging here.

christophertubbs (approved):
Looks good to me. Only found two nothing-burgers that aren't necessarily in scope or high priority

PR #447, opened 2023-10-03
Fixes and updates to GUI Docker image and service 

robertbartel:
Making a few changes, some to correct errors seen attempting to run things from scratch.

- Moving GUI Docker image to be based on Rocky Linux 9.1
- Explicitly listing _numpy_ as a project requirement in _pip_ requirements file used by GUI Docker image
- Adding flags to `pip` commands in GUI Docker image to avoid caching and reduce image size
- Fixing GUI service configuration to use correct path in Postgres-related Docker secret

### DIFF
--- a/docker/nwm_gui/app_server/Dockerfile
+++ b/docker/nwm_gui/app_server/Dockerfile
@@ -4,10 +4,18 @@
 ARG docker_internal_registry
 FROM ${docker_internal_registry}/dmod-py-sources as sources
 
-FROM python:3.8-slim as base
-# Update the aptitude package listing so that packages may be loaded and installed
-# Install python binaries to make pip installs a little smoother
-RUN apt-get update && apt-get install -y python3-django netcat
+FROM rockylinux:9.1
+
+# Do this separately at the beginning to get some caching help
+RUN dnf update -y && \
+    dnf install -y python3-pip git \
+    && ln -s /usr/bin/python3 /usr/bin/python
+
+RUN dnf install -y 'dnf-command(config-manager)' \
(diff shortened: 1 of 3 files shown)


docker/nwm_gui/app_server/Dockerfile
```diff
@@ -4,18 +4,26 @@
 ARG docker_internal_registry
 FROM ${docker_internal_registry}/dmod-py-sources as sources
 
-FROM python:3.8-slim as base
-# Update the aptitude package listing so that packages may be loaded and installed
-# Install python binaries to make pip installs a little smoother
-RUN apt-get update && apt-get install -y python3-django netcat
+FROM rockylinux:9.1
```
christophertubbs, line 7: Reasonable to be out of scope here, but the above + the `dnf` commands may work best as a common base image.

robertbartel: Yeah, I've had similar thoughts.  We need a fairly significant overhaul to the way we arrange, tag, and manage our images, and that's already on the road map.  For now, we get some of the same benefits mostly for free from the build cache.

robertbartel:
> The only thing I'm waffling on is that `--no-cache-dir` is a good line to use to force an update

@christophertubbs, I added those to help reduce the size of the images by not caching things, not as a means to force updates.  If I am misusing it - i.e., if it only prevents _checking_ the cache, and still results in writing to it (or somewhere that isn't cleaned up) - let me know.  I'm going to re-read the docs before I merge this to try to confirm the behavior.

PR #452, opened 2023-10-12
Evaluation Documentation

christophertubbs:
Documentation for all aspects of evaluations


aaraney:
Looking at this now.

python/lib/metrics/README.md
```diff
@@ -1,5 +1,273 @@
...
+```python
+        for metric in self.__metrics:  # type: Metric
+            self.__communicators.info(f"Calling {metric.name}", verbosity=Verbosity.LOUD, publish=True)
+            ...
+```
+
+This means that the `info` event will be triggered on each held `Communicator`, but only on those set to handle 
+messages of a verbosity of `LOUD` or greater, and to call the `write` event after doing so.
+
+Say I have three communicators:
```
aaraney, line 225: It would be super nice to see how these three example `Communicators` are created / configured.

python/lib/evaluations/dmod/evaluations/specification/README.md
```diff
@@ -0,0 +1,1338 @@
...
+![Specification for where to get the actual data for evaluation](../../../images/dmod.evaluations.specification.data.DataSourceSpecification.png)
+
+The [Data Source Specification](data.py) is the block of configuration responsible for loading a set of data
+that will be joined with others for evaluation. Loading data by using instructions provided by a `DataSourceSpecification`
+will yield a Data Frame that will be ready for manipulation.
+
+The most important aspects defined by a `DataSourceSpecification` are:
+
+1. What fields to load
+2. What the data is measured in or how to find out
```
aaraney, line 175: ~Are `udunits` supported? Or what is/are the expected unit type formats?~

python/lib/evaluations/dmod/evaluations/specification/README.md
```diff
@@ -0,0 +1,1338 @@
...
+        },
+        {
+            "template_name": "Median"
+        }
+    ]
+}
+```
+
+<a id="UnitDefinition"></a>
+## Unit Definition
```
aaraney, line 796: Is there a list of supported units / a dependency (i.e. `pint`) that you can point to in this section?

aaraney (changes requested):
This is great, @christophertubbs! I left a few comments, several of which are minor grammatical errors and others are optional suggestions.

python/lib/metrics/README.md
```diff
@@ -1,5 +1,273 @@
-# About
-Python package for utilities related to forcing, meta, and other modeling-related data for NWM MaaS.
+# Distributed Model on Demand - Metrics
 
-# Structure
-Structure has been modified from original to have inner duplicate directory in order to comply with general Python packaging structure.  This facilitates executing tests in a variety of different scenarios, including running integration tests on a local machine using the included test script and/or running unit tests directly within an IDE.
+The Distributed Model on Demand (DMOD) metrics package is a python library dedicated exclusively 
+to describing available functionality and performing mathematical operations upon provided data.
```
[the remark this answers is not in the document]
christophertubbs: `dmod.evaluations` is really the pipeline guy. `dmod.metrics` literally does nothing but provide some inspection utilities (these are the currently supported metrics and this is what they do) and actually scoring given dataframes. The primary 98% use case for it is building up the `ScoringScheme` object and passing dataframes into it.

aaraney (approved):
Thanks for working through the changes with me! Merge at will

PR #456, opened 2023-10-31
Updating DMOD Install Process and Docs

robertbartel:
- Updating the `INSTALL.md` file and other documentation related to setting up DMOD
- Updating scripts and config file templates to reduce the (required) manual steps to get DMOD up and running
- Adding new helper scripts, template files, and template file content in service of the above 
- A few miscellaneous fixes to issues within the setup process (e.g., Cython version for t-route in `ngen` image build)


doc/DEPENDENCIES.md
```diff
@@ -7,25 +7,28 @@ The DMOD project has two related sets of dependencies:
...
-| [Usage Dependencies](#usage-dependencies) | Same as above                               |                                                    |
-| Python                                    | \>=3.8.x                                    |                                                    |
-| Python Packages                           | See [requirements.txt](../requirements.txt) | Recommend installing in Python virtual environment |
+| Dependency                                | Constraints                                 |                                 Notes                                  |
+|-------------------------------------------|:--------------------------------------------|:----------------------------------------------------------------------:|
+| [Usage Dependencies](#usage-dependencies) | Same as above                               |                                                                        |
+| Python                                    | \>=3.8.x                                    |                                                                        |
```
christophertubbs, line 23: Might want to bump that up to 3.9.x - we've got 3.9 _somewhere_ in the pipeline (don't remember where). It's probably the github action.

robertbartel: The (perhaps unofficial) rule of thumb has been for DMOD to support all the versions of Python that are currently, officially supported, which includes 3.8 for now.  Given ecosystems like RHEL's that aren't as quick to move to new versions, I think this could make a difference in community adoption (which this particular PR is also geared toward).

I could have sworn we had a matrix in the Github Actions to support testing on a handful of different Python versions, including 3.8 and 3.9.  Maybe I'm mixing that up with somewhere else and we should add it here.

Regardless, if there is something specific for which we need to consider breaking with the current approach, then let's open a separate issue and formally consider it.

robertbartel: > I could have sworn we had a matrix in the Github Actions to support testing on a handful of different Python versions, including 3.8 and 3.9. Maybe I'm mixing that up with somewhere else and we should add it here.

Ah, looks like I was remembering what Nels had set up in the ngen-cal project.  We probably should adopt something similar.

christophertubbs:
Is this supposed to have all the DMOD client work?

christophertubbs:
All looks solid - I'll approve if it turns out that the client code wasn't an accidental branch leak.

robertbartel:
@christophertubbs , hmm, yeah that's probably a leak.  Let me figure out what I just did.

robertbartel:
@christophertubbs, good catch.  Thanks for that.  I'm not exactly sure how I did that, but I've removed the superfluous commits now.

christophertubbs:
If I had a nickel for every time I had a branch leak I'd have enough money for a pack of post-its to remind me not to cause a branch leak.

christophertubbs (approved):
Everything looks great

INSTALL.md
```diff
@@ -60,34 +101,53 @@ As illustrated, each service sub-directory must then contain two files:
...
 # Python Development
 
+## Installing Python and Path Verification
+
+As noted in the [dependencies](doc/DEPENDENCIES.md), Python, the Python development package, and Pip should be installed in the appropriate manner for your environment.  
+
+### Different Python Executable Names
+It is quite common for the Python executables to be named `python3`, `python311`, `pip3`, etc. in the *global* Python environment, instead of just `python` and `pip`.  This is accounted for when [creating a virtual environment](#using-a-virtual-environment), provided the right Python command was used to create it.  Users not doing this should consider setting up shell aliases, symlinks in their path, or some other mechanism to ensure `python` and `pip` execute the desired versions appropriately.
+
+> [!NOTE]
```
aaraney, line 136: Neat! I didnt know about these markdown directives.

Issue #459, opened 2023-11-01
Client API Incompatiliby with Utilized MinIO Version May Break Initial Setup


robertbartel:
As discussed in #242, we are using an old MinIO version, due to a deprecated feature related to backing storage and how it relates to the way DMOD is designed.  It turns out that newer versions of the MinIO client applications have a minor breaking change when used with the old service version.  

The break relates to [how access policies are set](https://min.io/docs/minio/linux/reference/minio-mc-admin/mc-admin-policy.html).  In short, the version used in DMOD requires the `admin policy set` command and does not support the `admin policy attach` command.  The reverse is the case for new versions of the client.

On a superficial level, this breaks the `minio_init.sh` helper script.  Updates are pending which will at least help exit cleanly in problem situations, but not to work around them.  

More importantly, it isn't possible to initialize the DMOD MinIO services - and thus, a DMOD deployment - using the currently configured version, without an older version of the MinIO client application.

aaraney:
Looks like https://github.com/minio/mc/releases/tag/RELEASE.2023-03-20T17-17-53Z is the breaking release.

aaraney:
Verified that [`RELEASE.2023-03-20T17-17-53Z`](https://github.com/minio/mc/releases/tag/RELEASE.2023-03-20T17-17-53Z) is the first version with the issue.

Based on how we check the version of the `mc` command in `scripts/minio_init.sh`, you can not use `go install github.com/minio/mc@RELEASE.2023-02-28T00-12-59Z` to get the last working version. You will need to either download a binary for that tag or clone the repo and run `make build`. Verified that this is does work, at least for me.

PR #460, opened 2023-11-01
Fixing Startup Issues with Several Services

robertbartel:
Primarily fixing issues with partitioner and request services, due to service client attributes (i.e., clients for *other* services) not having been updated to the new design using transport clients.

Also, making some changes to `minio_init.sh` to improve (but not fix; that'll have to wait for #459) how it deals with certain situations when there could be API incompatibility between the MinIO client (which the script uses) and the MinIO service.


christophertubbs (approved):
Everything looks solid and appears good to go.

I did spy a red flag that is far out of scope of this task, however.  The command line parser for services have waaaaay too many options (I eyeballed ~21) and the request service has several clusters of similar types of values. This is probably indicative that we need to find alternate ways to funnel settings to these applications to handle changes is a smoother fashion. This screams configuration settings for services and dynamic sets of configured service handlers in the request service to me.

PR #463, opened 2023-11-02
Update gh actions unit test

### LINKED ISSUE #457: Add Python version matrix in Actions workflow to test with different versions of Python
Use a matrix with current Github Actions for automated Python tests to support testing with multiple (and hopefully all) supported versions of Python.  An example of this is set up in the [ngen-cal package's Action workflow config](https://github.com/NOAA-OWP/ngen-cal/blob/8d46c053b4a6d553d0b6e2c4ac539b5c092a2a63/.github/workflows/ngen-cal.yaml).

aaraney:
- Add build python matrix. Tests now run on python `3.8`, `3.9`, `3.10`, and `3.11`.
- Add step to echo dependency versions. This is useful for debugging purposes.
- Fix TLS certs email todo. Certs now use phony `dmod_gh_action@fake.gov` email address.

closes #457

### DIFF
--- a/.github/workflows/testing_and_validation.yml
+++ b/.github/workflows/testing_and_validation.yml
@@ -104,6 +104,13 @@ jobs:
           cp -a ssl/local/*.pem ssl/requests/.
           cp -a ssl/local/*.pem ssl/scheduler/.
 
+      - name: Echo dependency versions
+        run: |
+          . dmod_venv/bin/activate
+          python3 -m pip freeze
+          deactivate
+        timeout-minutes: 1
+
       - name: Run Tests
         run: ./scripts/run_tests.sh --venv dmod_venv -v -d -srv
         timeout-minutes: 1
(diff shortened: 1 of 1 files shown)


aaraney:
Nice! So we have failures in `3.8` because of the use of `list[str]` instead of `typing.List[str]` as a type hint. See #464 for more detail.

aaraney:
Opened #467 to track a failure due to removal of the `ModelExecRequestClient` class.

aaraney:
Until https://github.com/NOAA-OWP/ngen-cal/pull/64 is merged, `pandas` will need to be built in each gh action run (unless we add a cache) for `3.10` and `3.11` b.c. `ngen-cal` requires an older version of `pandas` that is not pre-built for these versions.

aaraney:
https://github.com/NOAA-OWP/ngen-cal/pull/64 has been merged. This should be good to go now.

robertbartel (approved):

PR #470, opened 2023-11-15
bump min required version of to `aiohttp~=3.8`

aaraney:
Bump min required version of to `aiohttp~=3.8`. This should remove the warning in the dependency graph.


christophertubbs (approved):
👍

PR #471, opened 2023-11-16
Update ngen image with SQLite and GeoPackage hydrofabric support

robertbartel:
- Adding necessary flags within `docker/main/ngen/Dockerfile` to ensure ngen (and partitioner) is compiled with SQLite
- Updating `docker/main/ngen/Dockerfile` to use updated CMake cache variable names in appropriate commands, and updating associated `ARG` names to match

### DIFF
--- a/docker/main/ngen/Dockerfile
+++ b/docker/main/ngen/Dockerfile
@@ -46,7 +46,7 @@ ARG ROCKY_BASE_REQUIRED="sudo openssh openssh-server bash git which"
 
 ARG ROCKY_NGEN_DEPS_REQUIRED="sudo gcc gcc-c++ make cmake tar git gcc-gfortran libgfortran \
     python3 python3-devel python3-pip \
-    bzip2 expat expat-devel flex bison udunits2 udunits2-devel zlib-devel"
+    bzip2 expat expat-devel flex bison udunits2 udunits2-devel zlib-devel sqlite-devel sqlite-libs"
 # TODO: removed texinfo from list because it couldn't be found; make sure this doesn't lead to issues
 
 ARG BOOST_VERSION=1.82.0
@@ -76,6 +76,7 @@ ARG NGEN_NETCDF_ACTIVE="ON"
 ARG NGEN_ROUTING_ACTIVE="ON"
 ARG NGEN_UDUNITS_ACTIVE="ON"
 ARG NGEN_UDUNITS_QUIET="ON"
+ARG NGEN_WITH_SQLITE="ON"
 
 ARG BUILD_NGEN_SERIAL="true"
 ARG BUILD_NGEN_PARALLEL="true"
@@ -480,6 +481,7 @@ ARG NGEN_NETCDF_ACTIVE
 ARG NGEN_ROUTING_ACTIVE
 ARG NGEN_UDUNITS_ACTIVE
 ARG NGEN_UDUNITS_QUIET
+ARG NGEN_WITH_SQLITE
 
 ARG BUILD_NGEN_SERIAL
 ARG BUILD_NGEN_PARALLEL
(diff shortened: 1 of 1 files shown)


christophertubbs (approved):
This is out of scope, but a new ticket needs to be created to move the logic in the `RUN` statements into their own scripts - too much is in the docker file as it stands. Additionally, there are a bunch of hardcoded environment variables followed by bash logic that behaves based upon conditional logic from those hardcoded environment variables. If a line is required that states `REFRESH_BEFORE_BUILD=true`,  `if [ "${REFRESH_BEFORE_BUILD:-}" = "true" ]` becomes a truism. If there is room for configurability, .env files might help a little bit.

The changes for this task are good, but this file as a whole needs refactoring.

aaraney (approved):

robertbartel:
> This is out of scope, but a new ticket needs to be created to move the logic in the RUN statements into their own scripts - too much is in the docker file as it stands. 

Probably.  There is still a lot that needs to be done with this, both in design and organization.  With this suggestion in particular, I'd want to discuss it a little.  E.g., would we just be moving the mess to the directory structure?  

> Additionally, there are a bunch of hardcoded environment variables followed by bash logic that behaves based upon conditional logic from those hardcoded environment variables. If a line is required that states REFRESH_BEFORE_BUILD=true, if [ "${REFRESH_BEFORE_BUILD:-}" = "true" ] becomes a truism. If there is room for configurability, .env files might help a little bit.

There have been a lot of instances where I've added Dockerfile ARGs with a default value for future-proofing, which perhaps is excessive.  Some are added to simplify debugging (though in this case if I was doing that I probably should have set it to `false`).  

Regardless, there's a good bit more that still needs to be done to improve, extend, and specialize the image(s), which is very much on the radar.

christophertubbs:
> Probably. There is still a lot that needs to be done with this, both in design and organization. With this suggestion in particular, I'd want to discuss it a little. E.g., would we just be moving the mess to the directory structure?

Lines 511-606 and lines 641-671 would become their own scripts. Since they are in their own scripts, they won't rely on `\` at the end of lines and will leave room for more documentation, making things easier to read and understand.

robertbartel:
> Lines 511-606 and lines 641-671 would become their own scripts. Since they are in their own scripts, they won't rely on \ at the end of lines and will leave room for more documentation, making things easier to read and understand.

Maybe.  I can already think of times when I would have avoided errors modifying some Dockerfile by forgetting or removing a `\` or `&&`.   

That said - and I've by no means search exhaustively - after looking for best practices or possible pitfalls to the `RUN` instruction's use with scripts, I couldn't find much even talking about that approach.  The official [best practices for Dockerfiles](https://docs.docker.com/develop/develop-images/instructions/) page only seems to suggest the method of splitting up long or complex statements on multiple lines, at least with respect to `RUN` best practices.  

Interestingly, the `ENV` section *does* mention using scripts with `RUN` to control environmental variables.  I.e., the idea occurred to them, but didn't make the cut as a `RUN` best practice?

Put another way, is the problem really that we are chaining together so many statements, rather than doing something else?

Again, the images definitely need improvements,  and we should consider this.  It just seems a little conspicuous in its absence.

christophertubbs:
The reason you're probably not seeing a whole heck of a lot of "this is how you should do `RUN`s" or "maybe you should use scripts" is because both are viable options. I think some of the `RUN`s might be able to be split out, but it's the conditionals that throw a wrench in there. You'd want separate `RUN` statements to build little snapshots if there don't expect a lot of changes, but if you have something more complicated, you _do_ probably want that in one statement. Throw it in a script and you've got that snapshot as long as no commands prior change or that script changes. All `RUN` is, though, is an entry point to the interior shell of that state of the image, so calling `RUN chown ...` isn't really any different than `RUN ./script.sh`.

There are definitely pros and cons based on how you address it - many `RUN` statements will cause a longer initial build, introduce gotchas where more recent libraries aren't built into the image because instructions for intermediate images haven't changed, but makes subsequent builds _much_ faster and the `Dockerfile` a lot easier to read and track. Using a script will introduce a much faster **full** rebuild, allow you to follow all those steps whenever there's a minor change (I think you can probably change a sentinel variable value or something and it'll rerun the script upon build) to avoid the version gotcha as above, and make the `Dockerfile` much more concise, but it might make it harder to trace since you'll need to investigate another file and it'll reduce the number of intermediate images.

It's all a judgement call at this point.

Issue #472, opened 2023-12-18
Failure running t-route in ngen worker image


robertbartel:
Attempts to run framework-integrated t-route execution are failing.   Initially, these were encountering a segmentation fault.  After some experimental fix attempts, the errors changed first from to a signal 6, then to a signal 7, but t-route still does not run successfully.

The initial suspicion was a problem related to a [known NetCDF Python package issue](https://github.com/NOAA-OWP/t-route/issues/705), which is what early fix tries attempted to address (this may still be the root of what's going on).

hellkite500:
Can you make a `pip list` of the runtime python env?

robertbartel:
@hellkite500, sure:


<details>
    <summary>Output of <code>pip list</code> for ngen worker image</summary>

``` bash
[mpi@env4 ngen]$ pip list
WARNING: Ignoring invalid distribution -yarrow (/usr/local/lib64/python3.9/site-packages)
Package            Version
------------------ ------------
attrs              23.1.0
black              23.11.0
blosc2             2.3.2
bmipy              2.0.1
certifi            2023.11.17
cftime             1.6.3
click              8.1.7
click-plugins      1.1.1
cligj              0.7.2
Cython             3.0.6
dbus-python        1.2.18
Deprecated         1.2.14
fiona              1.9.5
geopandas          0.14.1
gpg                1.15.1
importlib-metadata 7.0.0
Jinja2             3.1.2
joblib             1.3.2
libcomps           0.1.18
MarkupSafe         2.1.3
msgpack            1.0.7
mypy-extensions    1.0.0
ndindex            1.7
netCDF4            1.6.3
numexpr            2.8.7
numpy              1.26.2
nwm-routing        0.0.0
packaging          23.2
pandas             2.1.4
pathspec           0.11.2
pip                23.0.1
platformdirs       4.1.0
py-cpuinfo         9.0.0
pyarrow            14.0.1
pyproj             3.6.1
python-dateutil    2.8.2
pytz               2023.3.post1
PyYAML             6.0.1
rpm                4.16.1.3
setuptools         53.0.0
shapely            2.0.2
six                1.15.0
systemd-python     234
tables             3.9.2
tomli              2.0.1
toolz              0.12.0
troute.network     0.0.0
troute.routing     0.0.0
typing_extensions  4.8.0
tzdata             2023.3
wheel              0.42.0
wrapt              1.16.0
xarray             2023.11.0
zipp               3.17.0
[...]

PR #474, opened 2024-01-04
Fix ngen image for building t-route

robertbartel:
Modifying image to get t-route built successfully, so that it is possible to run a NextGen execution that includes t-route routing.

Partially (but not completely) addresses #472; i.e., the image will successfully run serial `ngen` execution with at least some valid configurations of t-route.  Remaining problematic cases _appear_ to be not directly related to the worker image itself.

### DIFF
--- a/docker/main/ngen/Dockerfile
+++ b/docker/main/ngen/Dockerfile
@@ -430,6 +430,7 @@ USER ${USER}
 ##### Create intermediate Docker build stage for building framework in Rocky Linux environment
 FROM rocky-base as rocky_init_repo
 ARG REPO_URL
+ARG NGEN_BUILD_CONFIG_TYPE
 ARG BRANCH
 ARG COMMIT
 ARG WORKDIR
@@ -564,6 +565,7 @@ RUN cd ${WORKDIR}/ngen \
     && if [ "${BUILD_SLOTH}" == "true" ] ; then ./build_sub extern/sloth; fi \
     && if [ "${BUILD_NGEN_SERIAL}" == "true" ]; then \
         cmake -B cmake_build_serial -S . \
+        -DCMAKE_BUILD_TYPE=${NGEN_BUILD_CONFIG_TYPE} \
         -DNGEN_WITH_MPI:BOOL=OFF \
         -DNGEN_WITH_NETCDF:BOOL=${NGEN_WITH_NETCDF} \
         -DNGEN_WITH_BMI_C:BOOL=${NGEN_WITH_BMI_C} \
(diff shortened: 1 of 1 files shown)


docker/main/ngen/Dockerfile
```diff
@@ -374,32 +394,37 @@ RUN echo "export HYDRA_HOST_FILE=${HYDRA_HOST_FILE}" >> /etc/profile \
...
+RUN if [ "${NGEN_WITH_ROUTING}" == "ON" ]; then \
+      # Get subset of requirements needed first (including version specifiers if present) because of netCDF4 handling \
+      cat ${WORKDIR}/t-route-requirements.txt | grep -i numpy >> first_reqs.txt ; \
+      cat ${WORKDIR}/t-route-requirements.txt | grep -i pandas | grep -v geo >> first_reqs.txt ; \
+      cat ${WORKDIR}/t-route-requirements.txt | grep -i pyyaml >> first_reqs.txt ; \
+      cat ${WORKDIR}/t-route-requirements.txt | grep -i cython >> first_reqs.txt ; \
+      cat ${WORKDIR}/t-route-requirements.txt | grep -i pyarrow >> first_reqs.txt ; \
+      pip3 install --no-cache-dir -r first_reqs.txt ; \
+      rm first_reqs.txt ; \
+      # Then install a few more things, including a particular blosc2 version and an adjusted build of tables \
```
aaraney, line 417: > Looks like [`fiona`](https://pypi.org/project/fiona/1.9.5/) is the culprit. `geopandas` depends on `fiona` and `fiona` by default will install from `pip` with a pre-built `libgdal`, _if it can_. It seems like the issue is `fiona` does not have a distribution with pre-build `libgdal` for the particular target.
> 
> > Fiona has several [extension modules](https://docs.python.org/3/extending/extending.html) which link against libgdal. This complicates installation. Binary distributions (wheels) containing libgdal and its own dependencies are available from the Python Package Index and can be installed using pip.

[source](https://github.com/NOAA-OWP/DMOD/issues/475#issuecomment-1883705630)

```suggestion
      # Then install a few more things, including a particular blosc2 version and an adjusted build of tables \
      dnf install -y gdal-devel ; \
```

aaraney (changes requested):
Only needs a minor change to add support for arm. Looks good otherwise!

Aside, I don't know if we want to address this here or not, but on line 13 of the Dockerfile, `ARG NGEN_BUILD_CONFIG_TYPE="Release"` is not causing `ngen`, `ngen-serial`, or `ngen-parallel` to be built in Release mode. See the `rocky_build_ngen` build stage line ~565.

docker/main/ngen/Dockerfile
robertbartel: I fixed this differently, by including `gdal-devel` in the main step where the list of packages to be installed via `dnf` is defined.  It'll be installed now in an ancestor build stage of the one the suggestion would have been in.

robertbartel:
I addressed the `gdal-devel` item and fixed things with the value intended for `CMAKE_BUILD_TYPE`.  Should be ready for re-review now.

aaraney (approved):
Looks good to me! Thanks for making those changes. Merge at will.

Issue #468, opened 2023-11-02
`TestGeoPackageHydrofabric.test_uid_1_a` failure


aaraney:
Possibly related to #324.

Failure: https://github.com/NOAA-OWP/DMOD/actions/runs/6510147982/job/17683206387#step:10:319

Effected Code:

[`TestGeoPackageHydrofabric.test_uid_1_a`](https://github.com/NOAA-OWP/DMOD/blob/1f2faac423a90c04e74e87ea679c9469a0688d1b/python/lib/modeldata/dmod/test/test_geopackage_hydrofabric.py#L302) is failing.

aaraney:
So, a bit of background. We are using `pandas.util.hash_pandas_object` to generate a unique identifier for a geopackage hydrofabric. `test_uid_1_a` is failing b.c. the old hash does not match. This appears to be a regression in `pandas.util.hash_pandas_object`. This is the second time we've encoutered a regression in this function (see #324).

In doing some initial investigating, I came across [pandas documentation](https://pandas.pydata.org/pandas-docs/version/2.1/reference/index.html) that notes that the `pandas.util` package is considered private.

> The `pandas.core`, `pandas.compat`, and `pandas.util` top-level modules are PRIVATE. Stable functionality in such modules is not guaranteed.

With that in mind, we need to find an alternative stable alternative to compute a unique identifier that does not use `pandas.util`.

aaraney:
Issue is with `fiona`, not with `pandas`. Tested all `pandas >= 2.0.0` and all failed.

Issue is present in `fiona==1.9.5` but not `fiona==1.9.4.post1`. Thinking that is likely the call to [`fiona.listlayers`](https://github.com/NOAA-OWP/DMOD/blob/2faeb501f80c5ff189d5a7c62b25583b01c6f2a1/python/lib/modeldata/dmod/modeldata/hydrofabric/geopackage_hydrofabric.py#L346) that is now returning a different layer name list ordering.

aaraney:
Well, its never that easy. Looks like it is something else with `fiona`...

Either way, we should sort that layers list before hashing to create the unique identifier (#TODO @aaraney).

robertbartel:
I've confirmed in a quick test on two different machines (an Intel Mac and M2 Mac) that sorting the order of processing of the layers will produce consistent results when generating the unique id.  I'll leave it to you for the moment, @aaraney, but let me know if you want me to go ahead and open a PR.

aaraney:
[...]

PR #477, opened 2024-01-22
Fix hydrofabric unique id generation

robertbartel:
Fixing to ensure deterministic ordering of layers when building hash.  Should address #468.


robertbartel:
I did forget to actually fix the tests, and in fixing them I realized I had reversed things in my manual test of this fix on a second machine.  I.e., this by itself isn't giving consistent hashes across machines.  I'll leave as draft for now in case an additional adjustment ends up making this useful.

aaraney:
I just posted a few potential solutions to the issue in #468 if you want to have a look. In short, either pin `fiona>=1.9.5` or bring in the `pyogrio` library and use it as `geopandas`'s engine.

aaraney:
If all is well in #478, we can close this in favor.

PR #483, opened 2024-01-23
bug(py38): port `functools.cache` support

aaraney:
Fix `ImportError: cannot import name 'cache' from 'functools'` bug present in python3.8.

## Changes

- Fix `ImportError: cannot import name 'cache' from 'functools'` bug present in python3.8.


aaraney:
tests are failing b.c. of something that will be addressed in #424.

christophertubbs (approved):
Beautiful. I wouldn't have thought of the type guard and I like your use of the tuple check - I don't remember the way I usually do it, but this is better.

PR #485, opened 2024-01-23
Updated logging for the GUI and the EvaluationService

### LINKED ISSUE #484: The GUI may attempt to create an error log in the wrong location
The default logging location for errors in the DMOD GUI is found at "errors.log". This is fine at an application level, but when loading into a development console (such as through PyCharm), it will attempt to create the log at root and fail.

This has been solved with all the other log files with functions like:

```python
def get_maas_logging_filename() -> str:
    """
    Gets a suggested name for the core application log filename

    Controlled by the optional `APPLICATION_LOG_PATH` environment variable

    Returns:
        Gets a suggested name for the core application log filename
    """
    maas_log_filename = os.environ.get('APPLICATION_LOG_PATH') \
                        or os.path.join(application_values.BASE_DIRECTORY, f'{DEFAULT_LOGGER_NAME}.log')

    if not maas_log_filename.endswith(".log"):
        maas_log_filename += ".log"

    return maas_log_filename
```

A similar function for the error log should be created as well.

The EvaluationService GUI should be checked for a similar issue. If it is found there that should also be fixed.

christophertubbs:
Fixed an annoying issue where error logs would be attempted to be created at `/` under certain scenarios (not at normal runtime). Also covered a base where log file names of '' are somehow loaded and passed off as legitimate filenames.

resolves #484

### DIFF
--- a/python/gui/maas_experiment/logging.py
+++ b/python/gui/maas_experiment/logging.py
@@ -239,10 +239,8 @@ def get_maas_logging_filename() -> str:
     Returns:
         Gets a suggested name for the core application log filename
     """
-    maas_log_filename = os.environ.get(
-        'APPLICATION_LOG_PATH',
-        os.path.join(application_values.BASE_DIRECTORY, f'{DEFAULT_LOGGER_NAME}.log')
-    )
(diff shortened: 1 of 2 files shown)


aaraney (approved):
Looks good to me!

PR #487, opened 2024-01-24
Fix `dmod.scheduler` test warnings

aaraney:
Just some chores. Switch from deprecated set by attribute to setters. This gets rid of warnings in tests.


christophertubbs:
You missed an entry in the `monitorservice`. I _believe_ it's calling `dmod.scheduler`, so I imagine it falls under this umbrella.

robertbartel (approved):

PR #492, opened 2024-01-24
Fix evaluation service function type hints 

### LINKED ISSUE #491: 3.8 tests failing due to evaluation service?
https://github.com/NOAA-OWP/DMOD/actions/runs/7646533924/job/20835651730#step:11:320

~Looks like this is failing because the evaluation service doesn't have tests.~

Was fail because of using an ellipse in a type hint.

aaraney:
closes #491.

### DIFF
--- a/python/services/evaluationservice/dmod/evaluationservice/evaluation_service/wrapper.py
+++ b/python/services/evaluationservice/dmod/evaluationservice/evaluation_service/wrapper.py
@@ -48,11 +48,19 @@ class WrapperResults:
         return iter(self.__multiple_values)
 
 
+_T = typing.TypeVar("_T")
+_V = typing.TypeVar("_V")
+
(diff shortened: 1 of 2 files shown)


aaraney:
still trying to figure out why they are failing.

aaraney:
Figured out what the issue is. `typing.Callable[[typing.Any, ...],` is not parsable in 3.8.

https://github.com/NOAA-OWP/DMOD/blob/a9f5440e1e0d72b72cb4077a82378e1a3827935c/python/services/evaluationservice/dmod/evaluationservice/evaluation_service/wrapper.py#L52

christophertubbs (approved):
Good fixes all around - didn't know you didn't have to declare the ParamSpec beforehand.

We might need to find a solid way to only use typing_extensions only when the item isn't supported by typing yet.

PR #493, opened 2024-01-25
Fix `RedisDockerSwarmMonitor` type hints

aaraney:
Fix `RedisDockerSwarmMonitor` type hints

## Changes

- Fix `RedisDockerSwarmMonitor` type hints


christophertubbs (approved):
I like the formatting.

Stylistic note: `typing.Optional` is not necessary when you're setting a default to `None`.

PR #505, opened 2024-01-25
test: add `env_vars` pytest.ini option

aaraney:
use `env_vars` key in `pytest.ini` to set environment variables that will be set before tests run.


christophertubbs (approved):
Good as far as I can tell

aaraney:
sweet, thanks, @christophertubbs!

PR #506, opened 2024-01-25
chore: add missing `dmod.core` dep

aaraney:
`dmod.core` is required by both metrics and evaluation but was not listed in `setup.py`. This simply adds those missing deps.


christophertubbs (approved):

aaraney:
:tada:, thanks, @christophertubbs!

PR #509, opened 2024-01-29
test: add integration gh action

aaraney:
Add github action that runs integration tests. ~At the moment, integration tests are run only on `python 3.10`. Its up for discussion what version or versions of python we run these tests on and the frequency at which we run the tests.~

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -1,0 +1,149 @@
+# Test changes that have been pushed to the master and dev branches
+
+name: Intergration Tests
+
+# Controls when the action will run.
+on:
+  push:
+    branches: [ master, dev ]
+  pull_request:
+    branches: [ master, dev ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # Run unit tests on Python
+  integration_test_latest:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.8', '3.9', '3.10', '3.11']
+
+    # Service containers to run with `runner-job`
+    services:
+      # Label used to access the service container
+      redis:
+        # Docker Hub image
+        image: bitnami/redis
+        # Set health checks to wait until redis has started
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+          -e REDIS_PASSWORD="DPXzqRqjhsXokOVQcPUqOJuzKePMsfUc"
+
+        ports:
+          # Maps port 6379 on service container to the host
+          - 19639:6379
+
+      minio:
+        image: bitnami/minio:2022.10.8
+        ports:
+          - 9000:9000
+          - 9001:9001
+        options: >-
+          -e MINIO_ROOT_USER="dmod"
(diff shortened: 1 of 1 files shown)


aaraney:
These should fail because our integration tests have some regressions.

christophertubbs:
Why is this only added to 3.10? Sure, it fails, but one for all and all for one, amiright? May as well blow up all of them if we can't get it functioning for one of them.

aaraney:
@christophertubbs, yeah I think that is a fair point. They don't take that long to run since we don't have too too many of them and we do a fair amount of caching. My thinking was it most important to test on the version of python that we have in production. Im not sure if using a build matrix in this context will work, but let me force push something up and see what happens.

christophertubbs (approved):
Everything seems to fail correctly, so I'm pretty happy with the changes

robertbartel:
@aaraney, I'm not sure yet, but this may need to be reverted.  

The integration tests weren't intended to be run in Github.  Since they were meant to test how things worked together, at least some of the tests depend on certain parts of a DMOD deployment being initialized that we wouldn't want to try to do in Action runners, like create some of the Docker networks or a running object store stack.

We might be able to pull things apart, but I think those end up being sophisticated mocks for unit tests rather than integration test.

aaraney:
> It's also possible I'm overloading and/or misusing the term here, but regardless, as they are now, the "integration" tests run by the current tools were only expected to be run locally.

I think that is a fair interpretation. For better or worse IMO the interpretation of "integration" tests probably varies from one person to the next more than unit or end2end testing. To me, integration testing tests a subsystem or subsystems. That could include things like networking, but I don't think its inherently required to be considered an integration testing.

My thinking was the benefits still greatly out weight the costs of running some form of our integration tests using Github actions. Right, we might not catch issues stemming from the interaction with our deployment environment and our integration tests, but nothing is stopping us from running them in both places. Just here, we are guaranteed that they are being run frequently. I guess I see this as a why not? If we can use github actions and [service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers) to get us 85% of the way there, why not?

christophertubbs:
We need integration tests that can 'prove' component A,B, and C can talk to an object store (minio) **and** key-value store/message queue (redis). Testing docker stuff is always good (especially since this... you know... is meant to manipulate containers), but docker doesn't matter if bare components can't integrate with the elements they need to integrate with. Docker is a means to an end - using a swarm gets us _exactly_ what we need in a deployment environment. We won't be able to test if the scheduler or whatever works here since it'd have to spin up new containers, but we need to ensure that everything else works.

Call them smoke tests, call them `Tubbs-is-dumb` tests; it doesn't matter. We need tests that verify that our products not only pass their own unit tests but can also work with crucial 3rd party tools. These tests do that.

I'm personally against reversion here. @robertbartel , if you feel that borders are being crossed then the scripts themselves need to be altered to take docker/containerization out of the equation, not removing the actions themselves. If the integration tests test component-3rd party cooperation _and_ docker, we need to separate docker tests into their own set.

We need to stay away from tests that are only meant to run locally - all those would end up being are "works on my machine" tests. I separate that idea from 'docker environment' tests, especially since this product will eventually need K8 environment tests (coming 2087!).

Just about the **_only_** reason I can think of to remove these tests is if they somehow cause us to exceed compute limits imposed by github itself.

PR #511, opened 2024-01-30
fix: backwards compatibility of `NWMRequest`

aaraney:
Fix backwards compatible handling of `NWMRequest` `config_data_id` initializer argument.

## Changes

- Fix backwards compatible handling of `NWMRequest` `config_data_id` initializer argument.


python/lib/communication/dmod/communication/maas_request/nwm/nwm_request.py
```diff
@@ -51,8 +51,9 @@ def __init__(
             super().__init__(**data)
         else:
             data["request_body"] = dict()
-            nwm_inner_request_body = {"config_data_id": config_data_id}
-            data["request_body"]["nwm"] = nwm_inner_request_body
+            data["request_body"]["nwm"] = data["model"]["nwm"]
```
christophertubbs, line 54: Either this or the unit tests are wrong - looks like `model` isn't coming in through `data`.

christophertubbs (changes requested):

python/lib/communication/dmod/communication/maas_request/nwm/nwm_request.py
aaraney: @christophertubbs, nice catch, I just fixed it.

christophertubbs (approved):
Looks good.

Stylistic opinion:

Using an approach like:

```python
if "request_body" not in data:
    if "model" in data:
        request_body = data["model"][self.get_model_name()]
    else:
        request_body = {}

    if config_data_id is not None:
        request_body['config_data_id'] = config_data_id

    data['request_body'] = {
        self.get_model_name(): request_body
    }

super().__init__(**data)
```

will have a single terminal point instead of two and _might_ be a little easier to step through and predict. That's not important here since it's not exactly performing the most convoluted operation, but it is something that caught my eye.

PR #461, opened 2023-11-01
Template Imports and Exports

christophertubbs:
Added the ability to import and export evaluation templates in the evaluation service and added stock templates available to all evaluation service instances.

PR #512, opened 2024-01-30
fix: `Dataset` `__eq__` and `set_manager`

aaraney:
Reintroduce `Dataset`'s `__eq__` method. This was lost during the pydantic refactor. The current behavior is incorrect for several reasons. The most glaring being `Dataset`'s must have the same `manager_uuid` field values.

## Additions

- `Dataset.set_manager` method. This replaces setting a manager instance as if it were a property.

## Changes

- Reintroduce `Dataset`'s `__eq__` method. This was lost during the pydantic refactor. The current behavior is incorrect for several reasons. The most glaring being `Dataset`'s must have the same `manager_uuid` field values.

## Changes

- Use `dmod.core.Dataset`'s `set_manager` api.


aaraney:
Pretty sure I missed one place where a setter should be used now.

python/lib/core/dmod/core/dataset.py
```diff
@@ -177,19 +177,42 @@ def __init__(self, manager: DatasetManager = None, **kwargs):
...
-            # pydantic will not validate this, so we need to check it
-            if not isinstance(value.uuid, UUID):
-                raise ValueError(f"Expected UUID got {type(value.uuid)}")
-            self.manager_uuid = value.uuid
+        Parameters
+        ----------
+        value: DatasetManager
+            The new value for ::attribute:`manager`.
+        """
+        self._manager = value
```
robertbartel, line 214: This change removes support for unsetting the manager.  We weren't entirely handling the case properly, but I'm still curious why remove it rather than fix it.   And are we sure that we don't need it?

Also, why not also renaming the backing attribute to `manager`?

robertbartel (commented):
There are a few things I have a few questions about.

python/lib/core/dmod/core/dataset.py
```diff
@@ -177,19 +177,42 @@ def __init__(self, manager: DatasetManager = None, **kwargs):
...
             self.manager_uuid = manager.uuid
 
+    def __eq__(self, other):
+        return (
+            isinstance(other, Dataset)
+            and self.access_location == other.access_location
+            and self.category == other.category
+            and self.created_on == other.created_on
+            and self.data_domain == other.data_domain
+            and self.dataset_type == other.dataset_type
```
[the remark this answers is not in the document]
aaraney: Thanks for bringing up these questions! I had similar questions myself. I went back in the [git history to the first point before the `pydantic` refactor work](https://github.com/NOAA-OWP/DMOD/blob/ac8b0474497b479cdedcb394b8263a1f21f42acb/python/lib/core/dmod/core/dataset.py#L108-L112) and copied and pasted the `__eq__` implementation from that point. I thought that would at least be a good starting point.

I am not sure if we should compare `uuid`'s or not? I lean toward yes, but I could hear the argument for no. Our idea of a dataset has slightly changed since pre-`pydantic` refactor as we now use, more or less, aggregate datasets for everything but forcing data for a simulation. I think that change in thinking was probably why they were not compared in the previous version of the code when the thinking was to determine suitable existing datasets that could be used to satisfy a given simulation request? Is that fair and what are your thoughts?

python/lib/core/dmod/core/dataset.py
aaraney: Ah nice catch! I shouldn't have removed that capability. I was only thinking you would ever want to set a manager, but it makes sense that you may want to disassociate a manager too.

> Also, why not also renaming the backing attribute to `manager`?

If we change back to `manager`, if you were to set through the `manager` property, you wouldn't update the `manager_uuid` property unless you went through the `set_manager` method. My thinking has been for properties where in the past we'd done more than just set the properties value, hide the property and provide a setter.

python/lib/core/dmod/core/dataset.py
```diff
@@ -168,21 +168,42 @@ def _serialize_datetime(self: "Dataset", value: Optional[datetime]) -> Optional[
...
+        """
+        Sets the ::class:`DatasetManager` and updates the
+        ::attribute:`manager_uuid` property on this instances.
+
+        Parameters
+        ----------
+        value: DatasetManager | None
+            The new value for ::attribute:`manager`.
+        """
         self._manager = value if isinstance(value, DatasetManager) else None
```
aaraney, line 207: Got lost after force push. https://github.com/NOAA-OWP/DMOD/pull/512#discussion_r1475079148

python/lib/core/dmod/core/dataset.py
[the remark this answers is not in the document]
robertbartel: > Our idea of a dataset has slightly changed since pre-pydantic refactor as we now use, more or less, aggregate datasets for everything but forcing data for a simulation.

I don't think our abstract idea has change; we just evolved another more practical data format.  

Here's what I suggest:
- `expires`: test both are in the future
  - disagreeing about how long data is good for doesn't imply disagreeing on whether your are talking about the same data, as long as everyone agrees the data is still good
- `derived_from`: optionally test for equality
  - different _explicit_ origins implies different data, but lack of information about origin does not
- `derivations`: ignore
  - doesn't really say anything about "this" data 
- `last_updated`: optionally test for equality
  - if one side says things have updated more recently than the other, chances are these are not truly talking about the same data
- `uuid`: ??? ... for now, optionally test for equality, but this needs it's own issue for follow-up
  - I thought more about this, and I'm less sure of what to do; does it make sense to have a universally unique identifier if it is not universally unique (i.e., if there are ever two non-equal objects with the same value, even if the value is `None`)?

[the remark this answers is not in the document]
aaraney: Is it okay if `expires` is None on both datasets?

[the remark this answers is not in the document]
robertbartel: Ah, yes, I should have clarified, but I think that's fine also.  `None` implies the dataset is not temporary.   Certainly if both are `None`, this fits with equality.  And I don't think disagreeing on whether the dataset is temporary should necessarily imply looking at different data (again, as long as the expire time hasn't passed yet).

python/lib/core/dmod/core/dataset.py
```diff
@@ -168,21 +168,54 @@ def _serialize_datetime(self: "Dataset", value: Optional[datetime]) -> Optional[
...
 
         if value is not None:
```
robertbartel, line 221: I think we need to do this to make sure `manager_uuid` also gets properly unset.

```suggestion
        if value is None:
            self.manager_uuid = None
        else:
```

robertbartel (changes requested):

python/lib/core/dmod/core/dataset.py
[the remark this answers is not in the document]
aaraney: Thanks for clarifying!

aaraney:
@robertbartel, can you give this one quick last glance? I think we are just about ready to merge.

aaraney:
Just rebased to master and force pushed after the merge of #461.

python/lib/core/dmod/core/dataset.py
aaraney: Thanks for pointing this out, @robertbartel. Just pushed up a fix. I slightly altered the possible behavior, but I think its for the best. LMK if you think otherwise.

aaraney: Hmmm, so what about the case when you are initializing from a deserialized json message that includes `manager_uuid`? This boils down to is the invariant that an instance's `manager` and `manager_uuid` fields are always either `None` or some valid value? Or can you have a `manager_uuid` field without a `manager` being set? I think the answer if for sure the latter.

aaraney: FWIW the previous behavior pre-pydantic was to leave the `manager_uuid` field if you set the manager with a `None` value (or any non-covariant `DatasetManager` type).

robertbartel: > Hmmm, so what about the case when you are initializing from a deserialized json message that includes manager_uuid? 

Assuming we keep the `manager_uuid` attribute (I'll come back to this), we need to support this case.  The motive for including the `manager_uuid` field is that we can't serialize the manager, so instead we serialize a way to identify the right manager. 

> This boils down to is the invariant that an instance's manager and manager_uuid fields are always either None or some valid value?

Not strictly always.  Practically, but not strictly.  E.g., not immediately after deserialization.

> FWIW the previous behavior pre-pydantic was to leave the manager_uuid field if you set the manager with a None value (or any non-covariant DatasetManager type).

Right, I thought earlier in the discussion for the PR that this was an oversight.  The only reason to support setting the manager to `None`, beyond when we are first initializing the object, would be to unset the manager.  And we aren't properly unsetting the manager if the _Dataset_, when serialized, still reflects said manager's uuid.  For now though, leave this as you have it; I'm starting to think it more a consequence of a problem in the design than an oversight.

I don't want us to get further caught up here, so I'll open a separate issue, but I think we should examine whether it is really appropriate/necessary/useful to serialize the manager's uuid with a dataset.  Manager instance uuids end up being dynamically generated each time, and the service only permits one manager per _DatasetType_ (i.e., there's only ever one `OBJECT_STORE` type manager at a time, which gets an arbitrary uuid, so if we know a _Dataset_ is `OBJECT_STORE`, maybe we don't need to and shouldn't keep a manager uuid separately from a manager reference).

robertbartel (approved):

PR #513, opened 2024-01-30
chore: fix how session_secret is passed in

aaraney:
Scheduler integration tests are failing bc of this.


christophertubbs (commented):
lgtm

Only problem is that these parts are expecting a non-null `model_request` value, so that's throwing an error. The `"session_secret"` value seems fine, so I think this counts as complete.

christophertubbs (approved):

christophertubbs (approved):

PR #516, opened 2024-02-01
test: account for nested directories

aaraney:
Account for nested directories in `modeldata` integration tests. These tests will continue to fail until #512 is merged.

### DIFF
--- a/python/lib/modeldata/dmod/test/it_object_store_dataset_manager.py
+++ b/python/lib/modeldata/dmod/test/it_object_store_dataset_manager.py
@@ -101,7 +101,7 @@ class IntegrationTestObjectStoreDatasetManager(unittest.TestCase):
     def tearDown(self) -> None:
         # Remove testing datasets slated to be cleaned up
         for dataset_name in self._datasets_to_cleanup:
-            for obj in self.minio_client.list_objects(dataset_name):
+            for obj in self.minio_client.list_objects(dataset_name, recursive=True):
                 self.minio_client.remove_object(dataset_name, obj.object_name)
             self.minio_client.remove_bucket(dataset_name)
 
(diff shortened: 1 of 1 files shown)


robertbartel (approved):
That's not how I thought the API worked, though I could be wrong.  I thought it flattened everything inside the bucket, and just prefixed a faux ancestor path as part of object names to emulate a directory structure.  But as long as you've confirmed this still operates as needed, I'm good with it (I don't think it would break the current test even if it was wrong because of how the test data is structured).  

I am having unrelated issues with getting these tests to work in my own setup, so just to be safe I am going to leave it to you to actually merge it in.  It might not hurt to rebase the stuff from #512 first so we can see tests passing, but I'll leave that to you.

aaraney:
> That's not how I thought the API worked, though I could be wrong. I thought it flattened everything inside the bucket, and just prefixed a faux ancestor path as part of object names to emulate a directory structure.

Yeah you are right that directories don't really exist. However, `minio` and `s3` preserve the idea of a directory by treating and objects that contain a `/` in their name as if they were in a directory when you perform operations.


https://min.io/docs/minio/linux/developers/python/API.html#list_objects
`recursive` | _bool_ | List recursively than directory structure emulation.
-- | -- | --

aaraney:
> It might not hurt to rebase the stuff from #512 first so we can see tests passing, but I'll leave that to you.

Yeah I think that is a great caveat. Just rebased and force pushed.

aaraney:
All the `modeldata` integration tests are [passing now](https://github.com/NOAA-OWP/DMOD/actions/runs/7786256363/job/21230666564?pr=516#step:12:41). Going to go ahead and merge this. Thanks again, @robertbartel!

Issue #519, opened 2024-02-02
Fix integration testing setup for modeldata package to not require deployment


robertbartel:
Update integration testing setup tools and config for the modeldata package to permit successful testing without requiring partial setup of a DMOD deployment.  At present, the `test_package.sh` script will fail, first complaining about no `resources.yaml` file, then about `deployx` plugin needing to be set up (and possibly would about more things after this).  

These items are specific to actually deploying DMOD, but it should be possible to run most if not all package tests in a development environment that does not include a DMOD deployment.  If this somehow is not the case for certain tests, they should be separated into a group for that purpose.

aaraney:
If we are to go all in on `pytest`, we could move to using dmod `namespace` and `package` level `conftest.py` files to define and provide what `pytest` calls [`fixures`](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html). Thing of these as the `setUp` and `tearDown` methods of a `unittest.TestCase` in a single function. You can scope a given fixture to run at the function, class, module, package or session scope. For example, to replicate the behavior of `test_package.sh` we write a `package` scoped `fixture` that brings and then down a redis instance for example.

PR #520, opened 2024-02-05
test: skip scheduler integration tests if in CI

aaraney:
`scheduler` integration tests require access to docker and several pre-built images. GH actions dont support accessing a docker daemon instance present on a runner. At the moment determining a work around to this issue is not critical, so for the moment we will just disable these tests in CI.

Related to #519.

### DIFF
--- a/python/lib/scheduler/dmod/test/it_RedisBackedJobManager.py
+++ b/python/lib/scheduler/dmod/test/it_RedisBackedJobManager.py
@@ -13,7 +13,18 @@ from uuid import UUID
 
 # TODO: also add another test for the job manager factory and its factory method
 
+try:
+    import pytest
+    import os
+    # https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
+    in_ci = os.environ.get("CI", False) == "true"
+    skip_in_ci = pytest.mark.skipif(in_ci, reason="Tests require access to Docker Daemon. Do not have access to Docker Daemon in GH Actions.")
+except ImportError:
+    def skip_in_ci(fn):
+        return fn
 
+
+@skip_in_ci
 class IntegrationTestRedisBackedJobManager(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


aaraney:
This is working as expected. See the [logs](https://github.com/NOAA-OWP/DMOD/actions/runs/7786969857/job/21233001662?pr=520#step:12:44). Looking into why other tests are failing.

python/lib/scheduler/dmod/test/it_RedisBackedJobManager.py
```diff
@@ -13,7 +13,18 @@
 
 # TODO: also add another test for the job manager factory and its factory method
 
-
+try:
+    import pytest
+    import os
+    # https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
+    in_ci = os.environ.get("CI", False) == "true"
```
robertbartel, line 20: Curious if the use of `False` versus `"false"` was deliberate (following some convention) or unintentional.  It's fine, I just spent an extra couple of minutes thinking about whether it could lead to any trouble.

robertbartel (approved):
I did have a minor question, but it's more about style.  This looks good.

python/lib/scheduler/dmod/test/it_RedisBackedJobManager.py
aaraney: My thinking was the only truthy case should be when the string `true` is set. In all other cases it should be `False`. Thoughts?

robertbartel: It's mostly just counterintuitive to me at first glance.  I'd probably have used "false" as the default value for the `get` call and compared a string to a string.  But this is perfectly correct.

aaraney:
Thanks, @robertbartel!

PR #530, opened 2024-03-01
Support for Additional OWP BMI Modules in Worker Image

### LINKED ISSUE #528: Add support for baselines and skill scores
Evaluations have no means of comparing against a persistence base line in order to tell if predicted values are better or worse than `x <time units>` ago, which is fairly standard. 

A new `Specification` object should be created that will dictate:

1. How much time to lag
2. How to transform the values within the lag, if at all (mean, max, etc)

PyDantic handles the `timedelta` data type as either seconds in ints or floats or strings formatted as `"[-][DD ][HH:MM]SS[.ffffff]"` or `"[±]P[DD]DT[HH]H[MM]M[SS]S"`. It may be worth considering adding another field for numeric units and just defaulting to seconds, so users can enter `1` as the lag and "hours" as the unit and it'll be one hour instead of a user having to enter `3600`. It might also be work considering adding support for some of the other resampling formats. Don't add support for `pandas` data offset strings since they don't map to the timedelta type that the new model will contain.

It might be worth having the name of the transformation be an enum to reduce magic strings.

Add a new field to `EvaluationSpecification` of:

```python
    baselines: typing.List[typing.Union[PersistenceSpecification, DatasourceSpecification]] = Field(
        default_factor=list,
        description="Specifications for how to provide reference data for metrics requiring reference time series, such as skill scores"
    )
```

When loading data for the evaluations, each item in that `baselines` list should load up its own data. If it's a `DataSourceSpecification`, that should load like normal. If it's a `PersistenceSpecification` (name pending) the values in the observations should be shifted. If there is a given method, it needs to be rolled and had that method applied to it. Only aggregation functions available to a rolling window are allowable. 

Calculating the baseline can look like:

```python
import pandas
from datetime import timedelta

frequency = timedelta(hours=24)

output = pandas.read_csv(
    "/path/to/file.csv",
    parse_dates=["valid_date"]
)
output.set_index("valid_date", drop=True, inplace=True)

method_name = None
if method_name in ("mean", "max"):
    baseline = output['value'].rolling(window=frequency)
    if method_name == "mean":
        baseline = baseline.mean()
    else:
        baseline = baseline.max()
else:
    baseline = output['value']
    
baseline = baseline.shift(freq=frequency)
```
Column name mapping may need to be figured out if baseline column names clash with observations or forecasts.

Add metrics that require a baseline/reference, such as the Brier Skill Score. A list may be found at: https://www.cawcr.gov.au/projects/verification/. These metrics should communicate an error message and report an error message if no baseline values were provided. There should be a metric value per baseline.

The formula for graded performance will need to be adjusted to account for the additional metrics called. If there are 5 baselines and there are two metrics called, a skill score named A with a weight of 10, a skill score named B with a weight of 5, and C with a weight of 5, the total weight of the three metrics isn't 20 - they are 10 * 5 + 5 * 5 + 5.

Metrics that those new metrics rely on should be implemented as normal metrics as well (such as Ranked Probability Score for the Ranked Probability Skill Score).

robertbartel:
Including shared libraries for the following OWP developed BMI modules within the ngen worker image, and its derivatives:

- SoilFreezeThaw
- SoilMoistureProfile
- LGAR-C
- Snow17
- Sac-SMA

These are available within the image under `/dmod/shared_libs/`.

This closes #528.

### DIFF
--- a/docker/main/ngen/Dockerfile
+++ b/docker/main/ngen/Dockerfile
@@ -37,6 +37,10 @@ ARG LGAR_C_REPO_URL="https://github.com/NOAA-OWP/LGAR-C.git"
 ARG LGAR_C_BRANCH="master"
 ARG LGAR_C_COMMIT
 
+ARG SAC_SMA_REPO_URL="https://github.com/NOAA-OWP/sac-sma.git"
+ARG SAC_SMA_BRANCH="master"
+ARG SAC_SMA_COMMIT
+
 #### Default arguments for required dependencies needed during various build stages
 # The Rocky-Linux-based "base" stage, rocky-base
 ARG ROCKY_BASE_REQUIRED="sudo openssh openssh-server bash git which"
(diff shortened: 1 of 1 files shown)


aaraney (approved):

Issue #531, opened 2024-03-06
Cannot create new dataset with CLI client


robertbartel:
Error appears related to data format selection, though no formats seem to be viable currently:

```
(venv) [root@9c86f61a34bd dmod-cli]# python -m dmod.client dataset create --format=NGEN_REALIZATION_CONFIG test_realization
usage: dmod.client dataset create [-h] [--paths UPLOAD_PATHS [UPLOAD_PATHS ...]] [--data-root DATA_ROOT] [--continuous-restriction [CONTINUOUS_RESTRICTIONS ...]]
                                  [--discrete-restriction [DISCRETE_RESTRICTIONS ...]]
                                  [--format {AORC_CSV,NETCDF_FORCING_CANONICAL,NETCDF_AORC_DEFAULT,NGEN_OUTPUT,NGEN_REALIZATION_CONFIG,NGEN_GEOJSON_HYDROFABRIC,NGEN_PARTITION_CONFIG,BMI_CONFIG,NWM_OUTPUT,NWM_CONFIG,NGEN_CAL_OUTPUT,NGEN_CAL_CONFIG,NGEN_JOB_COMPOSITE_CONFIG,T_ROUTE_CONFIG}]
                                  [--domain-json DOMAIN_FILE]
                                  name {config,forcing,hydrofabric,observation,output}
dmod.client dataset create: error: argument --format: invalid choice: <DataFormat.NGEN_REALIZATION_CONFIG: (4, {<StandardDatasetIndex.CATCHMENT_ID: (1, <class 'str'>, 'CATCHMENT_ID')>: None, <StandardDatasetIndex.TIME: (0, <class 'datetime.datetime'>, 'TIME')>: None, <StandardDatasetIndex.DATA_ID: (2, <class 'str'>, 'DATA_ID')>: None}, None, True)> (choose from 'AORC_CSV', 'NETCDF_FORCING_CANONICAL', 'NETCDF_AORC_DEFAULT', 'NGEN_OUTPUT', 'NGEN_REALIZATION_CONFIG', 'NGEN_GEOJSON_HYDROFABRIC', 'NGEN_PARTITION_CONFIG', 'BMI_CONFIG', 'NWM_OUTPUT', 'NWM_CONFIG', 'NGEN_CAL_OUTPUT', 'NGEN_CAL_CONFIG', 'NGEN_JOB_COMPOSITE_CONFIG', 'T_ROUTE_CONFIG')
```

[...]

PR #537, opened 2024-03-07
Improve handling of client dataset creation attempt without domain

robertbartel:
Improve the client's behavior when it attempts to create a dataset but no data domain is defined/provided.  Previously, an error was raised.  Now, a response object will be returned by the relevant function to the client and displayed, indicating that the attempt failed and giving some information on why.  E.g., attempting something like this:

```bash
python -m dmod.client dataset create --format NGEN_REALIZATION_CONFIG test_composite_1 config
```

will now yield something like this: 

```json
{
   "message" : "Could not inflate keyword params to object due to RuntimeError - Cannot create DataDomain without at least one finite continuous or discrete restriction",
   "reason" : "No Dataset Domain Provided",
   "success" : false
}
```

This relates to #531, although strictly speaking this was not a bug preventing valid dataset creation (only an obscuring of an invalid attempt to create one).


robertbartel:
FYI, @aaraney and @hellkite500, I've resolved the previous test failures by merging in recent upstream stuff.

python/lib/client/dmod/client/dmod_client.py
```diff
@@ -64,6 +64,62 @@ def __init__(self, client_config: ClientConfig, bypass_request_service: bool = F
 
         self._auth_client: AuthClient = AuthClient(transport_client=self._get_transport_client())
 
+    def _extract_dataset_domain(self, **kwargs) -> DataDomain:
```
aaraney, line 67: Thoughts on making this a `staticmethod` just to make it clear that it does not modify the state of the instance, but makes sense to have it here for cohesion reasons?

aaraney (commented):
Thanks, @robertbartel! Just one comment. Should be good to go after that.

python/lib/client/dmod/client/dmod_client.py
robertbartel: Hmm, it's an interesting thought exercise ... I didn't really intend for things to make it possible to call this outside an instance; I just wanted to tidy up the code a bit.  On the other hand, aside from implying no state changes, it's probably slightly more performant.

Rather than fall all the way down the rabbit hole, I went ahead and made the change.

aaraney (approved):
LGTM, thanks @robertbartel!

aaraney:
Merge at will

PR #540, opened 2024-03-11
feat: Add data to dataset via object with `read()` method

### LINKED ISSUE #319: Add abstract add_item and add_items methods to DatasetManager
Add abstract `add_item` and `add_items` methods to [`dmod.core.dataset.DatasetManager`](https://github.com/NOAA-OWP/DMOD/blob/4677879c07bc6534b4aa3fd434804b51dc579ace/python/lib/core/dmod/core/dataset.py#L607) and add implementation to `DatasetManager` subtypes.

## Current behavior

The current facility for adding data to a `DatasetManager` subtype is using the [`add_data`](https://github.com/NOAA-OWP/DMOD/blob/4677879c07bc6534b4aa3fd434804b51dc579ace/python/lib/core/dmod/core/dataset.py#L650) method. The current interface gives flexibility to subtypes to support adding a single file, multiple files, or streaming blobs of data that are then rejoined into a single file or multiple files.

## Expected behavior

For clarity and convenience, new abstract methods `add_item` and `add_items` should be added to the `DatasetManager`. These methods should be of the following form:

```python
     def add_item(
        self,
        dataset_name: str,
        item_name: Union[str, Path],
        data: bytes,
        is_temp: bool = False,
    ) -> bool:
        ...

     def add_items(
        self,
        dataset_name: str,
        item_names: Iterable[Union[str, Path]],
        data: Iterable[bytes],
        is_temp: bool = False
    ) -> bool:
        """ `item_names` and `data` iterables must yield the same number of items. """
        ...

# Optional additions
from typing import Protocol
     
class Reader(Protocol):
    def read(self, size: int = -1, /) -> bytes:
        """ EOF if empty b''. """


    def add_buff_item(
        self,
        dataset_name: str,
        item_name: Union[str, Path],
        data: Reader,
        length: int = -1,
        is_temp: bool = False,
    ) -> Tuple[bool, int]:
        """
        `data`'s `read` method can be called multiple times and must return empty bytes (i.e.  b'')
        as EOF.

        `length` -1 indicates unknown length.

        Return indicates if addition was successful and the number of bytes added.
        """
        ...

```

aaraney:
Enable adding data via `add_data()` to existing dataset using an object with a `read() -> bytes` method. The `data` argument now accepts either a filename or something that has a `read() -> bytes` method. Implementation added for `ObjectStoreDatasetManager`.

closes #319

## Additions

- `DatasetManager.add_data()`'s `data` argument now accepts either a filename or something that has a `read() -> bytes` method. Implementation added for `ObjectStoreDatasetManager`.


python/lib/modeldata/dmod/modeldata/data/object_store_manager.py
```diff
@@ -249,8 +250,15 @@ def add_data(self, dataset_name: str, dest: str, data: Optional[bytes] = None, s
                                                       retain_until_date=datetime.now() + timedelta(hours=1))
             else:
                 retention = None
-            result = self._client.put_object(bucket_name=dataset_name, data=io.BytesIO(data), length=len(data),
-                                             object_name=dest, retention=retention)
+            if isinstance(data, Reader):
+                # Use AWS S3 default part size of 5MiB
+                # https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html
+                part_size = 5 * 1024 * 1024
```
robertbartel, line 256: Might it make sense to have this be a defined constant within the _Reader_ interface?

robertbartel (changes requested):
By and large this looks really good.  I had a couple of comments on the changes themselves, but very small things.  We should bump the _dataservice_ dependency versions for these packages, though, to make sure they get used.  Also, I haven't looked at why yet, but there are some test failures.

python/lib/core/dmod/core/dataset.py
```diff
@@ -546,8 +548,8 @@ def add_data(self, dataset_name: str, dest: str, data: Optional[bytes] = None, s
             A path-like string specifying a location within the dataset (e.g., file, object, sub-URL) where the data
             should be added.
         data : Optional[bytes]
```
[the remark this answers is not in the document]
aaraney: Ah, great catch! I missed this one! Just fixed it.

python/lib/modeldata/dmod/modeldata/data/object_store_manager.py
aaraney: I don't think so. The concept of a `Reader` does not need to know about part size IMO. Part size, from my understanding, is used by minio and S3 to chunk up an upload. If we run into performance issues here with uploading larger files in the future, we can either accept a different variant of a `Reader` that also implements `__len__` so we can more appropriately select a `part_size`, _or_ just take in an optional `part_size` in the method signature. At that point, we might just want to create a different chunked upload method to handle that case.

aaraney:
Sweet, looks like the tests are all passing now.

robertbartel (approved):

aaraney:
Thanks, @robertbartel!

PR #541, opened 2024-03-11
Fix test discovery

aaraney:
This is required as of `pytest>=8.1.0`.
See https://github.com/pytest-dev/pytest/issues/11475 for more detail


aaraney:
This should address the issues we are seeing in #540.

christophertubbs (approved):
Change looks fine, but someone needs to fix the syntax error in the scheduler

PR #543, opened 2024-03-12
feat: `ServiceSettings` object for `dataservice` environment variables

aaraney:
Adds `ServiceSettings` object to capture `dataservice` environment variables. Refactored where `dataservice` previously pulled in environment variables using `os.getenv()`.

## Additions

- Adds `ServiceSettings` object to capture `dataservice` environment variables


robertbartel (commented):
I like the general idea of encapsulating the settings, but I'm not seeing how we could allow a user to control the settings value if they needed for a deployment.  Granted, we didn't previously have everything finished to support all that, but we at least should make sure that isn't harder to do later.   What do you have in mind for that?

aaraney:
@robertbartel, sorry I should have dropped a link to [`pydantic`'s setting management docs](https://docs.pydantic.dev/1.10/usage/settings/). In brief, a `BaseSettings` subtype reads from environment variables (and even `.env` or docker `secrets` if you set that up) and does the expected pydantic type coercion and checking.

> One of pydantic's most useful applications is settings management.
>
> If you create a model that inherits from BaseSettings, the model initialiser will attempt to determine the values of any fields not passed as keyword arguments by reading from the environment. (Default values will still be used if the matching environment variable is not set.)

python/services/dataservice/dmod/dataservice/service_settings.py
```diff
@@ -0,0 +1,19 @@
...
+    s3fs_url_host: Optional[str]
+    s3fs_url_port: Port = 9000
+    s3fs_vol_image_name: str = "127.0.0.1:5000/s3fs-volume-helper"
+    s3fs_vol_image_tag: str = "latest"
+    s3fs_plugin_alias: str = "s3fs"
+    s3fs_helper_network: str = "host"
+
+
+    class Config(BaseSettings.Config):
+        frozen = True
```
robertbartel, line 19: I don't know what Pydantic > v1 does by default, but should we explicitly include `case_sensitive = False` here?

robertbartel (dismissed):
I've got one non-essential question about whether a tweak might be beneficial, but it certainly isn't necessary.  Feel free to merge.

python/services/dataservice/dmod/dataservice/service_settings.py
aaraney: Yep, great idea! I would rather err on the side of being explicit to make our lives easier whenever we inevitably need to port.

aaraney: @robertbartel, just added the change, rebased, and force pushed.

robertbartel (approved):

PR #544, opened 2024-03-12
feat: add `DatasetManagerCollection`; refactor data service to use new abstraction

aaraney:
Add `DatasetManagerCollection` abstraction to encapsulate a collection of `DatasetManager` objects and their associated `DatasetType`. A primitive version of this already exists in the code, this PR simply abstracts and refactors the primitive version into a well defined type.

## Additions

- Add `DatasetManagerCollection` abstraction to encapsulate a collection of `DatasetManager` objects and their associated `DatasetType`.

### DIFF
--- a/python/services/dataservice/dmod/dataservice/initial_data_adder_impl.py
+++ b/python/services/dataservice/dmod/dataservice/initial_data_adder_impl.py
@@ -10,6 +10,8 @@ from ngen.config.realization import NgenRealization, Realization
 from pathlib import Path
 from typing import Dict, List, Optional, Union
 
+from .dataset_manager_collection import DatasetManagerCollection
+
 
 class FromFilesInitialDataAdder(InitialDataAdder):
     """
@@ -110,15 +112,15 @@ class FromPartialRealizationConfigAdder(InitialDataAdder):
     # TODO: centrally define this somewhere else
     _REAL_CONFIG_FILE_NAME = 'realization_config.json'
 
-    def __init__(self, job: Job, all_dataset_managers: Dict[DatasetType, DatasetManager], *args, **kwargs):
+    def __init__(self, job: Job, dataset_manager_collection: DatasetManagerCollection, *args, **kwargs):
         """
 
         Parameters
         ----------
         job : Job
             The job requiring a realization config dataset, which must be an ngen-related job.
-        all_dataset_managers : Dict[DatasetType, DatasetManager]
-            Dictionary of all dataset managers currently available to the data service, keyed by dataset type.
+        dataset_manager_collection: DatasetManagerCollection
+            Collection of DatasetManager objects and their associated DatasetType.
         args
         kwargs
         """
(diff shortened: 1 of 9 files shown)


aaraney:
Appears github is experiencing a partial outage for PRs. Will check back in the morning to see if the changes I pushed are synced and fixed the test failures.

> Update - We're continuing to investigate an elevated number of pull requests that are out of sync on page load.
Mar 13, 2024 - 00:47 UTC
Update - We're continuing to investigate an elevated number of pull requests that are out of sync on page load.
Mar 13, 2024 - 00:12 UTC
Update - We're seeing an elevated number of pull requests that are out of sync on page load.
Mar 12, 2024 - 23:39 UTC
Investigating - We are investigating reports of degraded performance for Pull Requests
Mar 12, 2024 - 23:39 UTC

aaraney:
Looks like this is working now

aaraney:
Rebases and resolved merge conflicts post #543

python/services/dataservice/dmod/dataservice/initial_data_adder_impl.py
```diff
@@ -110,15 +112,15 @@ class FromPartialRealizationConfigAdder(InitialDataAdder):
...
         """
 
         Parameters
         ----------
         job : Job
             The job requiring a realization config dataset, which must be an ngen-related job.
-        all_dataset_managers : Dict[DatasetType, DatasetManager]
-            Dictionary of all dataset managers currently available to the data service, keyed by dataset type.
+        dataset_manager_collection: DatasetManagerCollection
+            Collection of DatasetManager objects and their associated DatasetType.
```
robertbartel, line 123: Did you mean to generalize this param/attribute to no longer necessarily contain all managers?  That was my first impression, although I could see interpreting things either way. 

If "yes," should we add some more graceful/informative error handling if a necessary manager is left out (alternatively, should/do we make sure that all the right managers are included; this seems like it'd be harder to do unless we go back to the "all" case)?  Possibly a custom exception to represent this and better communicate it up the stack? 

If "no", should we try to do something more to ensure the passed param really contains all managers (e.g., singleton-ize the service manager and/or its _DatasetManagerCollection_)?

Assuming we feel like something more needs to be done along one of these lines, we don't have to do it now, as long as we track it with another issue.

robertbartel (changes requested):
Overall, very good.  There is one small correction I think is needed for some docstring. There's also one other spot I had some questions about, which may require more issues be opened for later.

python/services/dataservice/dmod/dataservice/initial_data_adder_impl.py
aaraney: > Did you mean to generalize this param/attribute to no longer necessarily contain all managers? That was my first impression, although I could see interpreting things either way.

Great question. No, I just renamed it so it would be named consistent across the codebase. 



> If "no", should we try to do something more to ensure the passed param really contains all managers (e.g., singleton-ize the service manager and/or its _DatasetManagerCollection_)?

Maybe? I think the service is self contained enough at the moment that it doesn't merit adding that level of complexity. If this were intended to be a library that would be a different story. I'm persuadable though. Thoughts?

robertbartel: That'a fair assessment.  I would like to have it be a little more clear if/when these need to be "all managers." but I don't want to hold this up.  I've opened #554 to take a closer look at things and add some documentation, but we can address that later.

robertbartel (approved):

robertbartel:
@aaraney I've approved, but it looks like this needs rebasing.

aaraney:
@robertbartel, just rebased and repushed. Will you give this a quick glance when you get a chance? Thanks!

robertbartel (approved):

PR #546, opened 2024-03-13
chore: unpin `pytest` in CI 

aaraney:
`pytest==8.1.1` was failing to discover tests correctly because of an erroneous `__init__.py` file in the source tree (#545). Remove the `pytest` version pin and add the new `consider_namespace_packages = true` `pytest` option.
 
related to #541 and #542

#545 should be merged before this


christophertubbs (approved):

aaraney:
woot woot! Thanks, @christophertubbs!

PR #548, opened 2024-03-13
Refactor dataservice

aaraney:
## Changes
- Extract `dataservice` tasks into separate classes.


python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -1025,4 +830,206 @@ async def perform_checks_for_job(self, job: Job) -> bool:
...
+
+    Returns
+    -------
+    Dict[UUID, DatasetUser]
+        Map of all ::class:`DatasetUser` from all associated managers of the service, keyed by the UUID of each.
+    """
+    return {uuid: mgr.get_dataset_user(uuid)
+            for _, mgr in managers.managers() for uuid in mgr.get_dataset_user_ids()}
+
+class Count:
```
christophertubbs, line 1018: Unless this is to be used in 47k places for a myriad of reasons, can you use a name that more signifies the why/how of its intention? Maybe something like `ActiveOperationTracker` or something? This is a good place to be obnoxious, pedantic, and/or petty about the name.

python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -1025,4 +830,206 @@ async def perform_checks_for_job(self, job: Job) -> bool:
...
+    """
+    Unbound Semaphore-like class for counting acquisitions and releases. Unlike a semaphore, `acquire()` nor `release()`
```
christophertubbs, line 1020: Can you include an `Examples:` tag to show its expected use as a member in another object? Just something to show what to expect.

christophertubbs (dismissed):

python/services/dataservice/dmod/dataservice/service.py
aaraney: For sure! Thanks for putting some thought into a more useful name!

aaraney:
@robertbartel, just rebased.

python/services/dataservice/dmod/dataservice/__main__.py
```diff
@@ -6,14 +6,131 @@
...
-from .service import ServiceManager
-from .service_settings import ServiceSettings
-from .dataset_manager_collection import DatasetManagerCollection
-from dmod.scheduler.job import DefaultJobUtilFactory
 from pathlib import Path
 from socket import gethostname
 
+from dmod.scheduler.job import DefaultJobUtilFactory
+from dmod.modeldata.data.filesystem_manager import FilesystemDatasetManager
+from dmod.modeldata.data.object_store_manager import ObjectStoreDatasetManager
```
robertbartel, line 14: I think I didn't do this originally because I wanted to reduce the number of modules in which these types got imported (though I don't remember _why_ I wanted to do that).  That probably isn't worth worrying about - it likely does more harm than good (if any) by cluttering what's in the _service.py_ file - but I'll mention it just as a sanity check.

python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -992,4 +852,206 @@ async def perform_checks_for_job(self, job: Job) -> bool:
...
+    -------
+    Dict[UUID, DatasetUser]
+        Map of all ::class:`DatasetUser` from all associated managers of the service, keyed by the UUID of each.
+    """
+    return {uuid: mgr.get_dataset_user(uuid)
+            for _, mgr in managers.managers() for uuid in mgr.get_dataset_user_ids()}
+
+class ActiveOperationTracker:
+    """
+    Unbound Semaphore-like class for counting acquisitions and releases. Unlike a semaphore, `acquire()` nor `release()`
```
robertbartel, line 1042: Feels like there are some words missing here.  Not sure this is right either, but, e.g., should it say:

> Unlike a semaphore, _neither_ `acquire()` nor `release()` block.

python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -992,4 +852,206 @@ async def perform_checks_for_job(self, job: Job) -> bool:
...
+    job_util : JobUtil
+        Access and update jobs
+    dataset_manager_collection : DatasetManagerCollection
+        Facilitates creating and accessing Datasets
+    docker_s3fs_helper : DockerS3FSPluginHelper
+        Facilitates initialize Docker volumes for jobs
+    data_derive_util : DataDeriveUtil
+        Facilitates deriving data and datasets
+    tracker : ActiveOperationTracker
+        Semaphore-like object for signaling task processing state
```
robertbartel, line 934: The name and description here are very vague, and it's hard to tell what this is tracking and why without referencing the previous code.  

If I'm not mistaken, all usage of the new _ActiveOperationTracker_ is to replace what previously was done with the service manager's `_is_fulfilling_requirements`  attribute.   Let's rename the param and the attribute and have the docstring say a little more, here and in the other classes/functions where this new type is used.

python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -992,4 +852,206 @@ async def perform_checks_for_job(self, job: Job) -> bool:
...
+        await self._manage_temp_datasets()
+
+    async def _manage_temp_datasets(self) -> NoReturn:
+        """
+        Async task for managing temporary datasets, including updating expire times and purging of expired datasets.
+        """
+        while True:
+            # Ensure that mark and sweep doesn't proceed while something is potentially to linking datasets
+            while self._tracker.value > 0:
+                await asyncio.sleep(10)
```
robertbartel, line 887: Should we encapsulate this in the _ActiveOperationTracker_ class also?

robertbartel (changes requested):
For the most part this looks good.  There are a handful of changes I think are needed, and then a couple of things I bring up just in case more discussion will help.

aaraney:
Thanks for the reviews! Ill fix this up and have it ready for re-review in the morning.

aaraney:
Just pushed up resolutions for the requested improvements. This is ready for re-review now, @robertbartel!

robertbartel (approved):

aaraney:
Thanks, @robertbartel!

PR #553, opened 2024-03-15
CLI Fix and Related Service Improvement For Dataset Creation

robertbartel:
- Fix problem with internal name of CLI param for data format
- Improve handling of errors on service side so that a response still gets generated and makes it back to the client, even when errors/exceptions happen
  - Adding special handling of failure due to a dataset name that doesn't comply with MinIO bucket name restrictions


aaraney (approved):
This looks good to me. This will impact #548, but I think it makes the most sense to merge this and I can rebase that PR and merge these changes there.

aaraney:
Let's not worry about bumping the `dataservice` version at this point since so much is in flight around refactoring it.

hellkite500 (approved):

Issue #555, opened 2024-03-19
DockerServiceParameters - Default arguments are mutable


christophertubbs:
The signature for the constructor for the `DockerServiceParameters` class is:

```python
class DockerServiceParameters():
    def __init__(self, image_tag: str = None, constraints: list = [], user: Optional[str] = None, hostname: str = None,
                 labels: dict = {}, serv_name: str = None, mounts: list = [], env_vars: Optional[Dict[str, str]] = None,
                 secrets: List[SecretReference] = [], cap_add: Optional[List[str]] = None):
```

The provided arguments of `[]` and `{}` provide common **shared** defaults. If the defaults are used then populated, every other instance that starts to use those defaults will share the same data within the scope of the process.

Change the implementation to something like:

```python
class DockerServiceParameters():
    def __init__(self, image_tag: str = None, constraints: list = [], user: Optional[str] = None, hostname: str = None,
                 labels: dict = {}, serv_name: str = None, mounts: list = [], env_vars: Optional[Dict[str, str]] = None,
                 secrets: List[SecretReference] = [], cap_add: Optional[List[str]] = None):

        if constraints is None:
            constraints = []
            
        if labels is None:
            labels = {}

        ...
```

and find similar instances to fix.

aaraney:
Fixed by #557

PR #557, opened 2024-03-20
Quick `dmod.scheduler` cleanup

christophertubbs:
Converted legacy 'format' string formatting to modern f-strings, cleaned up odd exception signatures, put a couple todos at the locations of logic errors, and clarified some formatting

Just cleaned up some low hanging fruit I saw in the scheduler while looking through some of the code.


python/lib/scheduler/dmod/scheduler/scheduler.py
```diff
@@ -345,6 +355,8 @@ def _ds_names_helper(cls, job: 'Job', worker_index: int, category: DataCategory,
             msg = "Attempting to start {} job {} with more than allowed max of {} required {} datasets."
             raise RuntimeError(msg.format(job.model_request.__class__.__name__, job.job_id, max_count, category))
         elif None in dataset_names is None:
```
christophertubbs, line 357: This is a bizarre condition. I don't know what it's supposed to do, but it's not gonna do it.

python/lib/scheduler/dmod/scheduler/scheduler.py
```diff
@@ -399,9 +411,10 @@ def _generate_docker_cmd_args(self, job: 'Job', worker_index: int) -> List[str]:
 
         # Finally, convert the args map to a list, with each "flag"/key immediately preceding its value
         args_as_list = []
-        for flag_key, param_val in docker_cmd_arg_map.values():
+        for flag_key, param_val in docker_cmd_arg_map.items():
```
christophertubbs, line 414: No clue how this ever worked - this would be trying to split the items in `[str(len(job.allocations)), self.build_host_list(job), str(job.job_id), str(worker_index)]` in two. Going with `items` yields the keys and values described in the comment.

python/lib/scheduler/dmod/scheduler/job/job_manager.py
```diff
@@ -532,31 +529,37 @@ def delete_job(self, job_id) -> bool:
...
-                except:
-                    # TODO: look at at least logging something if this happens; for now ...
-                    pass
-                return True
+                    return True
+                except BaseException as exception:
+                    most_recent_exception = exception                    
+        
         # If we get here, it means we failed the max allowed times, so bail
+        logging.error(f"The job with ID '{job_id}' could not be deleted", exc_info=most_recent_exception)
```
aaraney, line 562: Uhhh, you could have up to 5 errors here. Ideally we would report all of these exceptions as they could be unique, although not likely.

python/lib/scheduler/dmod/scheduler/job/job_manager.py
```diff
@@ -673,43 +683,52 @@ async def manage_job_processing(self):
...
                 result = self.release_allocations(job_with_allocations_to_release.job_id)
                 if not result.success:
                     logging.error(f"Failure deallocating {job_with_allocations_to_release.job_id} due to "
                                   f"'{result.reason}' (message was: `{result.message}`)")
 
-            for job_transitioning_phases in jobs_completed_phase:
-                # TODO: figure out what to do here; e.g., start output service after model_exec is done
-                pass
+            # TODO: figure out what to do here; e.g., start output service after model_exec is done
+            #jobs_completed_phase = organized_lists[2]
```
aaraney, line 694: I would just remove these lines instead of commenting them out. Its up to your discretion.

python/lib/scheduler/dmod/scheduler/scheduler.py
```diff
@@ -20,10 +20,10 @@
     from .job import Job
 
 
-class DockerServiceParameters():
-    def __init__(self, image_tag: str = None, constraints: list = [], user: Optional[str] = None, hostname: str = None,
-                 labels: dict = {}, serv_name: str = None, mounts: list = [], env_vars: Optional[Dict[str, str]] = None,
-                 secrets: List[SecretReference] = [], cap_add: Optional[List[str]] = None):
+class DockerServiceParameters:
```
aaraney, line 23: Honestly, why dont we just make this a `dataclass` since that is all its used for?

python/lib/scheduler/dmod/scheduler/scheduler.py
aaraney: Yeah, I think this is always `False`?

python/lib/scheduler/dmod/scheduler/scheduler.py
aaraney: Yeah, great catch! This was for sure a bug.

python/lib/scheduler/dmod/scheduler/scheduler.py
```diff
@@ -653,8 +674,8 @@ def start_job(self, job: 'Job') -> Tuple[bool, tuple]:
         """
         model = job.model_request.get_model_name()
 
-        if len([a for a in job.allocations if a.cpu_count <= 0]):
-            msg = "Cannot start job {}; found allocation without positive CPU count"
+        if not any(a for a in job.allocations if a.cpu_count <= 0):
```
aaraney, line 677: I think this is backwards?

aaraney: I think you just need to drop the `not`.

aaraney (changes requested):

aaraney (changes requested):
Thanks for making these improvements! There is one spot where I think the logic is backwards and a few minor comments that are up for debate.

python/lib/scheduler/dmod/scheduler/job/job_manager.py
```diff
@@ -532,31 +529,37 @@ def delete_job(self, job_id) -> bool:
         ValueError
             If no job exists with given job id.
         """
-        logging.debug("Deleting job {}".format(job_id))
+        if not isinstance(max_attempt_count, int) or max_attempt_count < 1:
```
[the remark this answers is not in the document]
christophertubbs: It's mostly a CYA habit that meshes ok with the following "is it an `int` and is it at least one?" check that happens immediately. **_IDEALLY_**, that'd come from an environment variable with a fallback since it falls in line with a system tuning setting.

I can go ahead and change it if you feel strongly about it.

python/lib/scheduler/dmod/scheduler/job/job_manager.py
christophertubbs: The cleanest way to handle this would be with an `ExceptionGroup`, but that's a 3.11 feature (i.e. not in our minimum version). As I see it, my options here are maintaining the error from the latest attempt or combine formatted messages from each error. That can get hairy if/when tracebacks/frames or printing become involved. Something smart is always nice, but that's a little out of scope of "let's slap something on there for an X% improvement". Should I add a TODO or something?

python/lib/scheduler/dmod/scheduler/job/job_manager.py
christophertubbs: I only left it there because it seemed like there was a root of an idea there. Do I know what it was? No clue, but there _is_ a hint of an intention, so I wasn't sure whether or not to remove it. I'm for leaving it at the moment, just so people can see it if/when they go to play with it later and understand "Hey, this was designed with the intention that some step occurs here"

python/lib/scheduler/dmod/scheduler/scheduler.py
christophertubbs: Not a bad idea, but I didn't want to get that deep down in the weeds in something I'm not super knowledgeable about. That's a tad out of scope.

python/lib/scheduler/dmod/scheduler/job/job_manager.py
aaraney: Its better than it was when we started, so I think it is fine to keep it as is.

Yeah, ive been temped in several places in the code base to bring in the [`exceptiongroup`](https://pypi.org/project/exceptiongroup/) back port of that feature. Its harder to handle without pattern matching though.

python/lib/scheduler/dmod/scheduler/job/job_manager.py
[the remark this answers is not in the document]
aaraney: Yeah, could you please change it? It makes it a lot easier to follow.

aaraney (approved):
Looks great, thanks @christophertubbs!

aaraney:
This fixed #555

PR #563, opened 2024-04-01
docs: add evaluation architecture diagram

aaraney:
Diagram was created by @christophertubbs, just uploading.


christophertubbs:
The diagram you added was the _goal_ diagram. The system isn't there yet as it cannot evaluate locations in parallel. Maybe include both?

christophertubbs (approved):
Right on

PR #565, opened 2024-04-02
Initial item/file domain detectors

robertbartel:
Adding initial set of implementations of classes that can scan a data item/file and detect the applicable _DataDomain_ for its data.

Note that this branch builds off that of #564.  As such, this PR should remain in draft status until #564 is merged.


python/lib/core/dmod/core/dataset.py
```diff
@@ -487,6 +488,252 @@ def unlink_to_dataset(self, dataset: Dataset) -> bool:
         return dataset.manager is not None and dataset.manager.unlink_user(user=self, dataset=dataset)
 
 
+DataItem = TypeVar('DataItem', bound=Union[bytes, RepeatableReader, Path])
+
+
+class AbstractDomainDetector(ABC):
+    """ Abstraction for something that will automatically detect a :class:`DataDomain` for some data. """
+
+    def __init__(self, *args, **kwargs):
```
aaraney, line 497: I recommend that we drop this in favor of subtypes defining their own initializer that ideally is the only initializer (no need to call `super().__init__()`). In doing so, subtypes can more freely specify their requirements and loosen the coupling between the super and subclass.

python/lib/core/dmod/core/dataset.py
```diff
@@ -487,6 +488,252 @@ def unlink_to_dataset(self, dataset: Dataset) -> bool:
...
+
+        Raises
+        ------
+        DmodRuntimeError
+            If it was not possible to properly detect the domain.
+        """
+        pass
+
+    @property
+    def data_domain(self) -> DataDomain:
```
aaraney, line 524: I get that this is a nicety, but imo there is nothing stopping a subclass from implementing this behavior in their implementation of `detect`. This also places the burden on subtypes to update `_data_domain` when `detect` is called. I think the design can be simplified by dropping this property and the `_data_domain` class level instance variable. This could result in slightly more code, but the behavior of a subtype will be closer to its declaration.

python/lib/core/dmod/core/dataset.py
```diff
@@ -487,6 +488,252 @@ def unlink_to_dataset(self, dataset: Dataset) -> bool:
...
+        Returns
+        -------
+        DataDomain
+            The domain detected by the instance.
+        """
+        if self._data_domain is None:
+            self._data_domain = self.detect()
+        return self._data_domain
+
+    def reset(self):
```
aaraney, line 537: Should all subtypes be required to implement this? Or can split this out someway?

python/lib/core/dmod/core/dataset.py
```diff
@@ -487,6 +488,252 @@ def unlink_to_dataset(self, dataset: Dataset) -> bool:
...
+        """ Private flag of whether data item is a :class:`Path` object (that points to a file, per other checks). """
+        self._item_name = self._item.name if self._is_item_file else item_name
+        """ Name for the item; in some situations, contains important constraint metadata (e.g. catchment name). """
+        self._decode_format = decode_format
+        """ A decoder format sometimes used when reading data item in order to get metadata. """
+        if self._is_item_file and self._item.is_dir():
+            raise ValueError(f"{self.__class__.__name__} can't initialize with a directory path as its data item")
+
+
+class UniversalItemDomainDetector(ItemDataDomainDetector):
```
aaraney, line 628: Inline with some of the comments in #566, if this functionality is needed, I recommend we refactor this functionality into some kind of factory that is defined close to where it is used.

aaraney (changes requested):

python/lib/core/dmod/core/dataset.py
robertbartel: Hmm ... I'm not really making use of it, but I had wanted a lazily initialized property for accessing the domain value, and for all the detector types to have this.  My concern was if a detector needed to be used repeatedly (for the same data) that it'd be a performance hit to have to re-detect things over and over.  Though at this point I'm not sure though how often, if ever, that scenario is actually going to come up.  

Is that - complications from introducing a common, inherited property implementation - specifically what you are worried about here, or is there something more?

python/lib/core/dmod/core/dataset.py
robertbartel: > I get that this is a nicety, but imo there is nothing stopping a subclass from implementing this behavior in their implementation of detect.

True.  I think my motivation is that there is nothing making them either, nor making all of them do it in a consistent manner (e.g., if you want to re-detect versus return what was detected already, can you do that, and is it done the same way?).  That seemed most easily solvable by doing some things in the superclass. 

> This also places the burden on subtypes to update `_data_domain` when `detect` is called. 

Actually the reverse is true.  Calling `detect` is intended _not_ to impact `_data_domain`.  This is why `data_domain` and `reset` are implemented:  to control (and return) the `_data_domain` value (though, as I mentioned above, I think I'm not actually using them at the moment).   But `detect` should run detection and return a new object each time.   

> I think the design can be simplified by dropping this property and the _data_domain class level instance variable. This could result in slightly more code, but the behavior of a subtype will be closer to its declaration.

Heh, we're in another thing where my style shows itself to not be especially Pythonic.  

I'd intended to leverage the superclass to abstract a concern from the subclass implementations (though if we stick with that, I do need to at least go back and use the property in places).   We could skip that, and just have the independent `detect` definitions.  Subclasses would be more free to handle themselves, but I think we'd get slightly less predictable and consistent implementations.  I'm a little worried that, given how we're using these in places like the universal detector, that inconsistency could eventually become a problem, even if it isn't right now.

python/lib/core/dmod/core/dataset.py
robertbartel: To implement `reset`?  No, if we are keeping the lazy property and it's attribute at the superclass level, then `reset` should implemented there also.  

If we aren't keeping the lazy property, then whether `reset` is appropriate or not depends on the details of each subclass.

python/lib/core/dmod/core/dataset.py
aaraney: > but I had wanted a lazily initialized property for accessing the domain value, and for all the detector types to have this. My concern was if a detector needed to be used repeatedly (for the same data) that it'd be a performance hit to have to re-detect things over and over.

Yeah, I like that functionality and in some if not all cases that is desirable. My thinking is that subclasses can handle caching / lazy loading in their `detect` implementation. It likely will result in a minor amount of code duplication or at least code that looks repetitive, however I think considering the tradeoffs of having to know the relationship between the parent class and the subclass or a few lines of repetitive code I would rather have as much of the behavior in one place. Another, at least to me, interesting philosophical question. What are your thoughts?

python/lib/core/dmod/core/dataset.py
aaraney: > Heh, we're in another thing where my style shows itself to not be especially Pythonic.

Haha, I don't think your style is _not_ Pythonic. I just tend to have an aversion to inheritance outside of using it strictly for enforcing an interface. Especially, when you start sharing private properties between inheritance levels. So, for sure a style thing!

> I'd intended to leverage the superclass to abstract a concern from the subclass implementations (though if we stick with that, I do need to at least go back and use the property in places). We could skip that, and just have the independent `detect` definitions. Subclasses would be more free to handle themselves, but I think we'd get slightly less predictable and consistent implementations. I'm a little worried that, given how we're using these in places like the universal detector, that inconsistency could eventually become a problem, even if it isn't right now.

First off, thanks for articulating your design so throughly! I agree that abstracting the "caching" of the detected `DataDomain` in a consistent way is valuable. I think we can achieve this behavior, in a similar fashion, without requiring the functionality be provided by the base class in a consistent way by using composition instead. This has the benefit of decoupling the "caching" behavior from the detectors and provides a separate pathway to add complexity to "caching" mechanisms if needed in the future without having to muck around with how each `AbstractDomainDetector` works. What that could look like in practice is:

```python
class CachedDomainDetector:
  def __init__(self, detector: AbstractDomainDetector):
    self._detector = detector
    self._data_domain = None

  @property
  def data_domain(self) -> DataDomain:
     if self._data_domain is None:
       self._data_domain = self._detector.detect()
     return self._data_domain

  def reset(self):
    self._data_domain = None
```

Alternatively, it could be a subclass of `AbstractDomainDetector`. So, something like:

```python
class CachedDomainDetector(AbstractDomainDetector):
  def __init__(self, detector: AbstractDomainDetector):
    self._detector = detector
    self._data_domain = None

  def detect(self, **kwargs) -> DataDomain:
     if self._data_domain is None:
       self._data_domain = self._detector.detect()
     return self._data_domain

  def reset(self):
    self._data_domain = None

```

In either case, we are able to move the base class prescribed behavior elsewhere and decouple the behavioral concepts. The trade off is now you have to know about whatever that caching wrapper type is, but I would argue that once you find it, its far easier to understand the implications of that abstraction rather than having to walk up the inheritance tree and see how each parent type handles the, in this case, `_data_domain` property.

python/lib/core/dmod/core/dataset.py
robertbartel: Yeah, see my comment [here](https://github.com/NOAA-OWP/DMOD/pull/566#discussion_r1552134681) that (towards the bottom of it) discusses why I don't think a factory is feasible.   

In short, a typical factory can't pick the right detector subclass to create without a data format, and the practical solution to getting the necessary data format is by using domain detector.  So this _is_ a sort of brute-force bootstrapping factory, it just also implements the interface at the same time, since detecting the necessary domain is a by-product of how it works.

python/lib/core/dmod/core/dataset.py
robertbartel: I went ahead and just took it out.  I'm still not certain there wouldn't be complications moving to a composition like this that aren't worth the trouble for something as straightforward as a lazy property, but we didn't actually need it.  And we can always add the composition in the future if we later decide to.

robertbartel:
@aaraney, I think all the changes we've discussed have been made.  I'm a little puzzled by the test failures.  I'm not seeing them locally, and I created a fresh 3.10 environment to test in just to be safe.   I'm curious if you see these on your side.  FWIW, I did clear the cache entries related to this PR, and it didn't seem to help.

robertbartel:
@aaraney, @christophertubbs, is there anything remaining for this PR that needs to be changed before it can be approved?

python/lib/modeldata/dmod/modeldata/data/item_domain_detector.py
```diff
@@ -0,0 +1,211 @@
...
+        DataDomain
+            The detected domain.
+
+        Raises
+        ------
+        DmodRuntimeError
+            If it was not possible to properly detect the domain.
+        """
+        # TODO: (later) probably isn't necessary to treat separately, but don't have a good way to test yet
+        if isinstance(self._item, ReadSeeker):
```
aaraney, line 144: Like your TODO says, I don't think we need this.

https://geopandas.org/en/stable/docs/reference/api/geopandas.read_file.html
> filename
str, path object or file-like object
Either the absolute or relative path to the file or URL to be opened, or any object with a read() method (such as an open file or StringIO)

python/lib/modeldata/dmod/modeldata/data/item_domain_detector.py
```diff
@@ -0,0 +1,211 @@
...
+            gpkg_data = self._item
+
+        if isinstance(kwargs.get('region'), str):
+            raw_str = kwargs['region'].strip().lower()
+            if raw_str == 'conus':
+                vpu = None
+                conus = True
+            else:
+                conus = False
+                pattern = re.compile('(vpu)([-_]?)(\d+)')
```
aaraney, line 157: It looks like some VPUs have a directional suffix:

`nextgen_01.gpkg`
`nextgen_02.gpkg`
`nextgen_03N.gpkg`
`nextgen_03S.gpkg`
`nextgen_03W.gpkg`
`nextgen_04.gpkg`
`nextgen_05.gpkg`
`nextgen_06.gpkg`
`nextgen_07.gpkg`
`nextgen_08.gpkg`
`nextgen_09.gpkg`
`nextgen_10L.gpkg`
`nextgen_10U.gpkg`
`nextgen_11.gpkg`
`nextgen_12.gpkg`
`nextgen_13.gpkg`
`nextgen_14.gpkg`
`nextgen_15.gpkg`
`nextgen_16.gpkg`
`nextgen_17.gpkg`
`nextgen_18.gpkg`

aaraney (commented):
I think we are getting really close to being at a point to merge this. I left a few small comments that will likely turn into `TODOs`. Let's briefly continue the discussion, I don't want to hold this up much longer.

python/lib/core/dmod/core/dataset.py
```diff
@@ -486,6 +487,335 @@ def unlink_to_dataset(self, dataset: Dataset) -> bool:
...
+class ItemDataDomainDetectorRegistry:
+    """ A singleton registry in which to track the subtypes of :class`ItemDataDomainDetector`. """
+
+    _instance = None
+
+    @classmethod
+    def get_instance(cls) -> ItemDataDomainDetectorRegistry:
+        """ Get the singleton registry instance. """
+        if cls._instance is None:
+            cls._instance = ItemDataDomainDetectorRegistry()
```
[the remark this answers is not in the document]
robertbartel: I'm pretty sure this no longer will matter.  In #566, I've thrown out the registry entirely.

I'd hoped to separate things into these multiple PRs to make the process more clear, but that doesn't seem to have succeeded.  Does it make sense to continue with this PR, given all the changes from it are also in #566, or just move over to that one?

python/lib/core/dmod/core/dataset.py
```diff
@@ -486,6 +487,335 @@ def unlink_to_dataset(self, dataset: Dataset) -> bool:
...
+        ----------
+        subclass
+
+        Raises
+        -------
+        DmodRuntimeError
+            If the given subclass was not already registered.
+        """
+        name = subclass.get_registration_name()
+        if name not in self._detectors:
```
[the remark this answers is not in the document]
robertbartel: Another one that probably no longer applies.  At this point, I don't entirely remember why I thought it important to expect a valid call to unregister implied the subclass was currently registered.

python/lib/core/dmod/core/dataset.py
```diff
@@ -486,6 +487,335 @@ def unlink_to_dataset(self, dataset: Dataset) -> bool:
...
+        Returns
+        -------
+        str
+            The registration name for this subclass.
+        """
+        return cls._registration_name if cls._registration_name else cls.__name__
+
+    def __init__(self, item: DataItem, item_name: Optional[str] = None, decode_format: str = 'utf-8', *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._item: DataItem = item
```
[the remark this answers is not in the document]
robertbartel: Hmm, fair point.  I think it can be taken out, so I'll make that change.

python/lib/core/dmod/core/dataset.py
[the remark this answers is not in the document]
aaraney: Sorry, I should have looked at changes made there. Yeah, let's not hold things up here.

aaraney (approved):
I'm seeing that a lot of the comments I've made here have already been addressed in #566. Let's move this along and merge this.

PR #568, opened 2024-04-03
Transition configuration of data service to env vars / `.env` file / docker `secrets` 

aaraney:
Drop `dataservice` CLI argument parsing and replace with a mechanism for configuring the service with environment variables, a `.env` file (does not need to be `export`ed), docker `secrets`, or a combination of the three. The primary goal of this change is to simplify service configuration and ultimately deployment. As an added benefit, the use of pydantic to validate and coerce configuration options will also slightly harden the service. 

See `README.md` in change set for a listing of configuration options, their types, and descriptions. 

## Additions

- Service is now configured using either, environment variables, a `.env` file, docker style `secrets` (stored at `/run/secrets/`), or a combination of the three. Input variable names are case insensitive and are sourced in the following priority:
  1. environment variables
  2. variables loaded from a dotenv (`.env`) file.
  3. variables loaded from the secrets directory.
  4. default field values

## Removals

- Remove support for providing service configuration as CLI arguments.
- `FilesystemDatasetManager` support has been dropped. The feature was only partially functional and can easily be added in the future once in feature parity with the existing `ObjectStoreDatasetManager`.

## Changes

- Docker deployment and entrypoint script updated to function with feature set.


python/services/dataservice/dmod/dataservice/service_settings.py
```diff
@@ -1,20 +1,203 @@
...
+    """Change the base directory when using SSL certificate and key files with default names"""
+
+    object_store_host: str = "minio-proxy"
+    """Set hostname for connection to object store"""
+    object_store_port: Port = Port(9000)
+    """Set port for connection to object store"""
+
+    # TODO: consider making this a pydantic.SecretStr
+    object_store_username: str
+    """Set name of the Docker secret containing the object store user access key"""
```
robertbartel, line 41: We probably shouldn't engineer something at this level in a way that is quite so tailored to our use of Docker.  I suggest we rename this and the other similar items to suggest that they are really the names of files (at least as far as the app is concerned), not only to be more clear, but so we can have separate attributes for the values.  We may not always acquire the value in this manner in the future.

aaraney: > We probably shouldn't engineer something at this level in a way that is quite so tailored to our use of Docker

Are you primarily suggesting that I should change the doc string to remove the reference to Docker? I did not make it clear in the source documentation that any defined field can be sourced programmatically (when initializing the object), from an environment variable, from a `.env` file, or from a file with the fields name stored in `/run/secrets` (following Docker's convention). I don't think we should do this now, but if we _need_ to configure _where_ secrets are sourced, there is a straight forward pathway to do that.

python/services/dataservice/dmod/dataservice/__main__.py
```diff
@@ -1,100 +1,74 @@
...
+import sys
+import typing
+
+import pydantic
+
 logging.basicConfig(
     level=logging.INFO,
     format="%(asctime)s,%(msecs)d %(levelname)s: %(message)s",
-    datefmt="%H:%M:%S"
+    datefmt="%H:%M:%S",
```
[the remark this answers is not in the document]
aaraney: Nope! Trailing commas in argument lists are totally fine. I use the python formatter [`black`](https://black.readthedocs.io/en/stable/index.html) and that is how it formats multiline sequence of arguments.

docker/main/docker-deploy.yml
```diff
@@ -159,26 +159,40 @@ services:
...
     environment:
       - SERVICE_PACKAGE_NAME=${PYTHON_PACKAGE_NAME_DATA_SERVICE:-dmod.dataservice}
-      - LISTEN_PORT=${DOCKER_DATASERVICE_CONTAINER_PORT:-3015}
-      - SERVICE_SSL_DIR=/ssl/data-service
-      - USER_SECRET_NAME=object_store_exec_user_name
-      - PASSWD_SECRET_NAME=object_store_exec_user_passwd
+      - PORT=${DOCKER_DATASERVICE_CONTAINER_PORT:-3015}
+      - CERT_PATH=/ssl/data-service/certificate.pem
+      - KEY_PATH=/ssl/data-service/privkey.pem
+      - SSL_DIR=/ssl/data-service
```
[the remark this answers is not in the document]
aaraney: Great question! I was thinking the same thing when I was refactoring this. I was not sure (and didn't lookup) if you can provide root and intermediate x509 certs in separate files along side the private and public key? Meaning, I didn't know if we dropped support for the `SSL_DIR` if this would required bundling the root and intermediate certs with the public key cert and if there were reasons we wouldnt want to do that.

If this isn't an issue, I would love get rid of it!

docker/main/docker-deploy.yml
```diff
@@ -159,26 +159,40 @@ services:
...
       - OBJECT_STORE_HOST=${DMOD_DATA_SERVICE_OBJ_STORE_HOST:-minio-proxy}
-      #- REDIS_PORT=
-      #- REDIS_USER=
+      - OBJECT_STORE_USERNAME=object_store_exec_user_name
+      - OBJECT_STORE_PASSWD=object_store_exec_user_passwd
```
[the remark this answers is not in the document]
aaraney: Yeah, I think that is totally fair. Thanks for mentioning this! Are you thinking adding `exec` into the configuration field names (e.g. `OBJECT_STORE_EXEC_USERNAME`)?

docker/main/docker-deploy.yml
```diff
@@ -159,26 +159,40 @@ services:
...
+      # - PYCHARM_REMOTE_DEBUG_EGG='/pydevd-pycharm.egg'
+      # - REMOTE_DEBUG_HOST=${PYCHARM_REMOTE_DEBUG_SERVER_HOST:-host.docker.internal}
+      # - REMOTE_DEBUG_PORT=${PYCHARM_REMOTE_DEBUG_SERVER_PORT_DATA_SERVICE:-55871}
       #- VENV_DIR=${DOCKER_REQUESTS_CONTAINER_VENV_DIR:-}
+      # debugging flags
+      #- UPDATED_PACKAGES_DIR=${UPDATED_PACKAGES_CONTAINER_DIR:?Updated packages directory not set, make sure this should be active}
     working_dir: /code
     ports:
       - "${DOCKER_DATASERVICE_HOST_PORT:-3015}:${DOCKER_DATASERVICE_CONTAINER_PORT:-3015}"
+    secrets:
```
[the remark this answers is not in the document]
aaraney: Great catch!

docker/main/docker-deploy.yml
```diff
@@ -243,6 +257,10 @@ secrets:
     file: ${DOCKER_REDIS_SECRET_FILE:?Variable DOCKER_REDIS_SECRET_FILE unset in env; no path to myredis password Docker secret file}
   # Have to create these here from the same backing files, since the stack prefix causes issue with sharing secrets
   # across stacks
+  object_store_passwd:
```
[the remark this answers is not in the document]
aaraney: Okay! Sounds good to me.

aaraney:
@robertbartel, I think ive addressed all of you comments and this is ready for re-review. I would like reorder / squash the commits Ive added to address your comments before we merge, but for review sake I did not want to do that yet since it might make reviewing more difficult.

python/services/dataservice/dmod/dataservice/service_settings.py
robertbartel: The docstring was part of it.  My general concern was that if `object_store_username`/`object_store_exec_user_name` stored the name of a Docker secret, then using that attribute name would be misleading unless we assumed certain things about having Docker in the deployment.  

But from your response, it sounds like somehow  _ServiceSettings_  gets the content of secret files automatically as appropriate.  I'm guessing that it knows to puts those contents into `object_store_exec_user_name`, instead of the value directly provided, when the provided value is actually something in the secrets dir.  Alternatively, if the value within an `.env` file for `object_store_exec_user_name` is just a literal value for the user name, the _ServiceSettings_ object can tell the difference and just use the literal value.  I just don't quite see how that works.  

As long as that is how it works, then this is fine (apart from the elsewhere discussed `exec` stuff for the name).  And that would explain why the environment variables themselves aren't named to reflect that they are Docker secret names.   Though the way things all fit together is perhaps a bit obfuscated and could use some clarifications (certainly it was with the previous docstring, though I see you've already changed that).

aaraney: Yep, you nailed it. That is pretty much exactly how it works. I tried to emphasize that in the README. Im sure ive missed the mark somewhere. Can you skim it and let me know holes i've left? I want to make sure it is clear how to configure the service.

python/services/dataservice/README.md
```diff
@@ -2,4 +2,74 @@
...
+      ```
+      PORT=8080
+      PORT='8080'
+      PORT="8080"
+      ```
+For more in-depth information, see these resources:
+- https://docs.docker.com/compose/environment-variables/env-file/#syntax
+- https://github.com/theskumar/python-dotenv?tab=readme-ov-file#file-format
+
+A secret's filename must be the name of a configuration variable.
```
robertbartel, line 32: I feel a little more strongly about this one, but only a little.

```suggestion
A secret's filename must match the name of a configuration variable (case insensitive).
```

python/services/dataservice/README.md
```diff
@@ -2,4 +2,74 @@
...
 # Structure
-Structure has been modified from original to have inner duplicate directory in order to comply with general Python packaging structure.  This facilitates executing tests in a variety of different scenarios, including running integration tests on a local machine using the included test script and/or running unit tests directly within an IDE.
+Structure has been modified from original to have inner duplicate directory in order to comply with general Python packaging structure.
+This facilitates executing tests in a variety of different scenarios, including running integration tests on a local machine using the included test script and/or running unit tests directly within an IDE.
+
+## Configuration
+
+Provide listed service configuration variables as either environment variables, in a '.env' file, as a `secret` in `/run/secrets`, or as a mixture of the aforementioned options.
+
+Configuration option names are case insensitive.
```
robertbartel, line 12: I think this would be helpful, but not necessary:

```suggestion
Configuration variable names are case insensitive.
```

python/services/dataservice/dmod/dataservice/service_settings.py
```diff
@@ -1,20 +1,203 @@
...
+- each line represents a key-value pair. values can optionally be quoted.
+  for example:
+      PORT=8080
+      PORT='8080'
+      PORT="8080"
+For more in-depth information, see these resources:
+- https://docs.docker.com/compose/environment-variables/env-file/#syntax
+- https://github.com/theskumar/python-dotenv?tab=readme-ov-file#file-format
+
+A secret's filename must be the name of a configuration variable.
```
robertbartel, line 98: Similar to the above:

```suggestion
A secret's filename must match the name of a configuration variable (case insensitive).
```

robertbartel (dismissed):
I think this is probably fine, but I did have a couple of minor inline suggestions.  We can discuss further, skip, or commit and re-approve.

My original issue was that it's not hard to accidentally conflate "configuration variable" with "environment variable" if you aren't paying close attention **AND** I got thrown off by the inclusion of object-store-related, secret-backed config variables as entries in the service config `environment` section in _docker-deploy.yml_.   With the latter out of the way, I think things are made clear enough.

python/services/dataservice/README.md
aaraney: Yeah, I went back and forth when writing this. We usually say a CLI tools command line options, not its command line variables. I think it depends on your background which term you prefer. I am going to leave it as is. Thanks for the careful read and feedback!

aaraney:
Thanks for the careful read, @robertbartel! I made several of the changes you suggested. I am going to squash / reorder some of the commits and force push just to make this clearer and request a re-review. None of the content from this point forward will change, just the git metadata.

PR #575, opened 2024-04-11
feat: port data service to FastApi; add POST endpoint for uploading data to a dataset

### LINKED ISSUE #538: Enable upload and download of DMOD dataset data via REST

aaraney:
closes #538 

The goal of this PR is to port the current data service's web server backend from the python library `websockets` to the python web server framework `FastApi` _and_ create a POST endpoint for uploading data to an existing dataset. The "new" stack still utilizes the `websockets` library under the hood. However, now an [`asgi`](https://asgi.readthedocs.io/en/latest/introduction.html) server layer has been added between the `websocket` / `http` protocol layer and the application layer. This gives us flexibility if we need to move away from the `websockets` library for example, as now we can choose whatever `asgi` server is best suited to our needs. `asgi` servers typically allow selecting which protocol specific library (e.g. `websockets`, `wsproto`) is used to handle either the `http` or `websockets` protocols. For simplicity, i've selected the `uvicorn` `asgi` server which  supports the `websockets` library. We should discuss if this selection is appropriate and discuss other options before merging this.

There is some nuance in _what_ has changed in the service's network stack. The previous service used from the bottom up, python's `asycio.loop.create_server` (`websockets` calls this) to handle  `IP`, `TCP`, and `TLS`, then the `websockets` library to serialize and deserialize `HTTP`, and `websockets`. Our application code uses the abstractions provided by the `websockets` library to communicate. 

The new stack decouples the underlying websockets transport library by utilizing a `asgi` server and framework. As mentioned previously, the `asgi` server, `uvicorn` is now used. `uvicorn` is responsible for calling `asycio.loop.create_server` which still handles `IP`, `TCP`, and `TLS` however, `uvicorn` utilizes libraries like `h11` and `websockets` for example which implement the `asycio.Protocol`'s for handling HTTP and websockets protocols respectfully. `uvicorn` enables communication via the `asgi` specification, thus decoupling it from the underlying communication mechanisms (i.e. `websockets`). 

`FastApi` is an `asgi` _framework_ which sits on top of the `starlette` `asgi` framework. `starlette` does most of the work and implementing a `http` multiplexer and middleware registry among other things. `FastApi` provides a lot of niceties through its heavy integration with `pydantic` which enables request and response validation + openapi documentation. The service's http / websocket communication is now implemented using `FastApi`. In theory, this will enable us to pivot to other `asgi` servers / protocol libraries is necessary. It will also give us the added benefit of having openapi documentation which could be used to generate client code.

Potentially weird things you might come across when reviewing this PR:
- `fastapi.Depends` - `FastApi` provides a dependency injection feature for providing dependencies to a given handler. The classic example of this is a database connection. If you are familiar with `pytest` fixtures, these will look and feel similar in _use_ but not definition. It is a common pattern within the `FastApi` app community to decorate a function with `@lru_cache` to create effectively a singleton (if the function has no arguments). Ive followed this pattern for several of the dependencies required by the websocket handler.
- The `lifespan` function. This will run before the service is started _and_ while the service is shutting down. If you are familiar with `pytest` fixtures, they work almost identically. Everything before the yield is run before server startup and everything after yield is run during shutdown. As it sits, there is a lot going on in that function. It definitely become a place for the procedural code. Im personally fine with it staying that way, but I would love to hear what you all think and ways you would rather see the cruft managed.

## Additions

- `/add_object` POST endpoint. This enables adding an object to and existing dataset. Adding an object that already exists will overwrite it.

## Changes

- `FastApi` used instead of `websockets` library
- Connection can only be made to the websocket using `/` path. Previously, any valid URL path could be used. In the future, I would like to change this to `/ws`. However, this is out of scope for this change.

## TODO:
- Update service README

### DIFF
--- a/.github/workflows/testing_and_validation.yml
+++ b/.github/workflows/testing_and_validation.yml
@@ -44,7 +44,7 @@ jobs:
           . dmod_venv/bin/activate
           pip install --upgrade pip
           deactivate 
-          ./scripts/update_package.sh --venv dmod_venv --no-service-packages --dependencies
+          ./scripts/update_package.sh --venv dmod_venv --dependencies
 
       - name: Cache Package Checksums
         id: cache-packages-md5
--- a/docker/main/docker-deploy.yml
+++ b/docker/main/docker-deploy.yml
@@ -184,6 +184,8 @@ services:
       #- VENV_DIR=${DOCKER_REQUESTS_CONTAINER_VENV_DIR:-}
       # debugging flags
       #- UPDATED_PACKAGES_DIR=${UPDATED_PACKAGES_CONTAINER_DIR:?Updated packages directory not set, make sure this should be active}
+    # uncomment to override entry point; intended for development use ONLY
+    # entrypoint: ["python", "-m", "uvicorn", "dmod.dataservice.rest_service:app", "--port=8080", "--host=0.0.0.0"]
     working_dir: /code
     ports:
       - "${DOCKER_DATASERVICE_HOST_PORT:-3015}:${DOCKER_DATASERVICE_CONTAINER_PORT:-3015}"
--- a/python/services/dataservice/dmod/dataservice/__main__.py
+++ b/python/services/dataservice/dmod/dataservice/__main__.py
@@ -1,8 +1,5 @@
 import logging
 import sys
-import typing
-
-import pydantic
 
 logging.basicConfig(
     level=logging.INFO,
@@ -25,11 +22,11 @@ from .service import (
     ServiceManager,
     TempDataTaskManager,
 )
-from .service_settings import DebugSettings, ServiceSettings, field_usage
+from .service_settings import ServiceSettings, debug_settings, service_settings
 
 
 def main():
-    settings = _handle_service_settings()
+    settings = service_settings()
 
     if settings.pycharm_debug:
         _setup_remote_debugging()
@@ -110,16 +107,8 @@ def main():
     service_manager.run()
 
 
-def _handle_service_settings() -> ServiceSettings:
-    return _handle_settings(ServiceSettings)
-
-
-def _handle_debug_settings() -> DebugSettings:
-    return _handle_settings(DebugSettings)
-
-
 def _setup_remote_debugging():
-    settings = _handle_debug_settings()
+    settings = debug_settings()
     logging.info("Preparing remote debugging connection for data service.")
     if not settings.pycharm_remote_debug_egg.exists():
         print(
@@ -148,23 +137,6 @@ def _setup_remote_debugging():
             ),
             file=sys.stderr,
         )
-
-
-_T = typing.TypeVar("_T", bound=pydantic.BaseSettings)
-
-
-def _handle_settings(cls: typing.Type[_T], **kwargs) -> _T:
-    try:
-        return cls(**kwargs)
-    except pydantic.ValidationError as e:
-        usage: typing.List[str] = []
-        for err in e.errors():
-            field_name: str = err["loc"][0]
-            msg = f"{field_usage(ServiceSettings, field_name)}\nError: {err['msg']}\n"
-            usage.append(msg)
-        error_msg = "\n".join(usage)
-        print(f"Service configuration failure: {error_msg}", file=sys.stderr)
-        sys.exit(1)
 
 
 def _init_object_store_dataset_manager(
--- a/python/services/dataservice/dmod/dataservice/dataset_manager_collection.py
+++ b/python/services/dataservice/dmod/dataservice/dataset_manager_collection.py
@@ -12,6 +12,7 @@ class DatasetManagerCollection:
     For each DatasetType, there must be only one DatasetManager.
     A given DatasetManager instance may be associated with multiple DatasetTypes.
     """
+    # TODO: consider how `__eq__` should be implemented (#598)
     _managers: Dict[DatasetType, DatasetManager] = dataclasses.field(
         default_factory=dict, init=False
     )
--- a/python/services/dataservice/dmod/dataservice/dataset_manager_collection.py
+++ b/python/services/dataservice/dmod/dataservice/dataset_manager_collection.py
@@ -16,6 +16,9 @@ class DatasetManagerCollection:
         default_factory=dict, init=False
     )
 
+    def __hash__(self) -> int:
+        return id(self)
+
     def manager(self, dataset_type: DatasetType) -> DatasetManager:
         """
         Return the manager for the given dataset type.
--- a/docker/main/dataservice/entrypoint.sh
+++ b/docker/main/dataservice/entrypoint.sh
@@ -23,4 +23,4 @@ if [ -d ${UPDATED_PACKAGES_DIR:=/updated_packages} ]; then
     fi
 fi
 
-python -m ${SERVICE_PACKAGE_NAME:?}
+python -m uvicorn dmod.dataservice.rest_service:app --host "${HOST:?HOST not set}"--port "${PORT:?PORT not set}" --ssl-keyfile="${KEY_PATH?:KEY_PATH not set}" --ssl-certfile="${CERT_PATH:?CERT_PATH not set}"
--- a/python/services/dataservice/dmod/dataservice/_injectable.py
+++ b/python/services/dataservice/dmod/dataservice/_injectable.py
@@ -1,0 +1,67 @@
+"""
+Functions that injectable via FastAPI's dependency injection system.
+"""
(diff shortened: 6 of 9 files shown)


aaraney:
Several service settings can and should be removed as a part of this PR too.

python/services/dataservice/dmod/dataservice/dataset_manager_collection.py
```diff
@@ -16,6 +16,9 @@ class DatasetManagerCollection:
         default_factory=dict, init=False
     )
 
+    def __hash__(self) -> int:
+        return id(self)
```
robertbartel, line 20: Maybe there are some implications of _dataclass_ I'm forgetting/missing, but don't we essentially get the same functionality automatically for a user-defined class (i.e., equality and hash equality only for the same object)?

And beyond whether we get that for free, is that really what we want here?  I.e., why wouldn't we want any object with an equal dictionary of managers to test/hash equally?  

Those probably are questions for use cases that, in practice, aren't going to matter much, so the PR doesn't need to get held up while you add full hashing and equality (i.e. a TODO will suffice, if even that much is needed).  But since you seemed to go out of your way to add this, I wanted to at least ask about it.

robertbartel (commented):
I've done a first pass, and for the most part this looks good.  There are a few things I think may need to be fixed that weren't quite rebased correctly, and a few others I had questions about.

.github/workflows/testing_and_validation.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           . dmod_venv/bin/activate
           pip install --upgrade pip
           deactivate 
-          ./scripts/update_package.sh --venv dmod_venv --no-service-packages --dependencies
+          ./scripts/update_package.sh --venv dmod_venv --dependencies
```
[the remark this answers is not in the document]
aaraney: Yeah, I wasn't sure why that flag was present either. That ended up being the reason why the tests were failing (the service pkgs were not being updated). I'm happy to do either. I figured that if I saw `update_package.sh` I would think that it would install all the packages, not just the `lib` packages, so I just left it like this. Im indifferent.

python/services/dataservice/dmod/dataservice/dataset_manager_collection.py
aaraney: Thanks for pointing this out, @robertbartel! I should have added a docstring here to clarify the intent. A `dataclass` will only create a `__hash__` if it is "safe" to do so. That is to say that the internal fields of the `dataclass` are immutable. 

From [`dataclasses`](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) documentation:

>__hash__() is used by built-in [hash()](https://docs.python.org/3/library/functions.html#hash), and when objects are added to hashed collections such as dictionaries and sets. Having a __hash__() implies that instances of the class are immutable. Mutability is a complicated property that depends on the programmer’s intent, the existence and behavior of __eq__(), and the values of the eq and frozen flags in the @dataclass decorator.
>
>By default, @dataclass will not implicitly add a [__hash__()](https://docs.python.org/3/reference/datamodel.html#object.__hash__) method unless it is safe to do so. Neither will it add or change an existing explicitly defined __hash__() method. Setting the class attribute __hash__ = None has a specific meaning to Python, as described in the __hash__() documentation.

`__hash__` will not be generated for this class since a `dict` is not hashable. However, for all intents and purposes, we use this around the application like a singleton. To enable passing the same instance around to each `fastapi` handler that needs it, instead of using a global variable, we can use `fastapi`s dependency injection system to inject an instance into a handler. To ensure that we are passing the same instance around and not creating new instances, we decorate the injectable function with `functools.lru_cache` to effectively return the same instance each time the injectable is executed. However, to make this work, `functools.lru_cache` requires that arguments are hashable. This means if any _other_ injectable function _requires_ our injectable `DatasetManagerCollection` and it also is decorated with `lru_cache`, then `DatasetManagerCollection` _needs_ to be hashable. So, the simplest way to do this was just return the `id` reference to the class instance.

To illustrate the problem in [code](https://github.com/NOAA-OWP/DMOD/blob/01aab671f41466a7757b6fe543e37e286374aeb4/python/services/dataservice/dmod/dataservice/_injectable.py#L15):

```python
@lru_cache
def dataset_manager_collection() -> DatasetManagerCollection:
    """
    Service singleton `DatasetManagerCollection`.
    All handlers or service level background tasks that need a `DatasetManagerCollection` _should_
    use this / depend on it via FastApi's dependency injection system.
    """
    return DatasetManagerCollection()


@lru_cache
def data_derive_util(
    dmc: Annotated[DatasetManagerCollection, Depends(dataset_manager_collection)]
) -> DataDeriveUtil:
    """
    Service singleton `DataDeriveUtil`.
    All handlers or service level background tasks that need a `DataDeriveUtil` _should_
    use this / depend on it via FastApi's dependency injection system.
    """
    return DataDeriveUtil(dataset_manager_collection=dmc)
```

`data_derive_util` depends on `dataset_manager_collection` as an argument _and_ is decorated with `lru_cache`, so the return value of `dataset_manager_collection` must be hashable.


FastApi mentions this in their [docs](https://fastapi.tiangolo.com/advanced/settings/?h=lru_cache#creating-the-settings-only-once-with-lru_cache) as a solution to passing around globals.

docker/main/docker-deploy.yml
```diff
@@ -251,6 +256,10 @@ secrets:
     file: ${DOCKER_REDIS_SECRET_FILE:?Variable DOCKER_REDIS_SECRET_FILE unset in env; no path to redis password Docker secret file}
   # Have to create these here from the same backing files, since the stack prefix causes issue with sharing secrets
   # across stacks
+  object_store_passwd:
```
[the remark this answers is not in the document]
aaraney: Great catch! Just dropped that erroneous commit, so we should be good. Can you double check me on that please?

python/services/dataservice/dmod/dataservice/__main__.py
```diff
@@ -150,23 +139,6 @@ def _setup_remote_debugging():
         )
 
 
-_T = typing.TypeVar("_T", bound=pydantic.BaseSettings)
-
-
-def _handle_settings(cls: typing.Type[_T], **kwargs) -> _T:
```
aaraney, line 156: This got moved to `service_settings.py`

python/services/dataservice/dmod/dataservice/dataset_manager_collection.py
robertbartel: > A `dataclass` will only create a `__hash__` if it is "safe" to do so.

Ah, gotcha.

> `__hash__` will not be generated for this class since a dict is not hashable.

Yeah, like I said before, we shouldn't hold this up, but I did some rudimentary testing, and I think it ends up being the case that two _DatasetManagerCollection_ objects could test as equal but have different hashes.  That is a slightly more practical, though still not-often-encounter violation of the intent for hashing.   So let's make sure we add a TODO to come back to this later.

aaraney: Ah, I see what you are saying. I wasn't thinking about `__eq__`. Are you suggesting that I overload `__eq__` like I have `__hash__`?

aaraney: With a pragmatic lens, we don't ever compare `DatasetManagerCollection`s. We could instead just raise not implemented. Thoughts?

.github/workflows/testing_and_validation.yml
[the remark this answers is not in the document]
robertbartel: No, as I think more about it, what you've done is probably what _needs_ to be done, and that not already being done was a flaw.  And we can fix the part of the flaw applicable here to the data service by fixing it more generally.

python/services/dataservice/dmod/dataservice/dataset_manager_collection.py
robertbartel: Let's defer the conversation, but put something in place that reminds us to come back it.  As you say, we don't compare those right now.  We might later.  So we should be careful and thorough in thinking how this should be, but time is better spent on other things right now.

aaraney: I am going to leave it as is for now an add a `TODO` with reference https://github.com/NOAA-OWP/DMOD/issues/598

aaraney: Added a `TODO` and reffed the related issue

robertbartel (approved):

PR #576, opened 2024-04-11
Set websocket close timeout to 5 seconds

### LINKED ISSUE #574: Set appropriate websocket `close_timeout`
As described in the websocket [documentation](https://websockets.readthedocs.io/en/stable/reference/asyncio/common.html#websockets.legacy.protocol.WebSocketCommonProtocol):

> `close_timeout` (Optional[[float](https://docs.python.org/3/library/functions.html#float)]) – Timeout for closing the connection in seconds. For legacy reasons, the actual timeout is 4 or 5 times larger.

Digging a little deeper, the `close_timeout` parameter was once named `timeout` (we also don't specify this either) and defaults to `10` seconds. If neither `timeout` nor `close_timeout` are provided, the default, now, `close_timeout` is 10 seconds. As the docs allude to, for "legacy" reasons this is, in practice at least 2x the timeout amount. 

Why does this matter? Per the [websocket rfc 6455](https://www.rfc-editor.org/rfc/rfc6455#section-7.1.2):

> Once an endpoint has both sent and received a Close control frame, that endpoint SHOULD _Close the WebSocket Connection_ as defined in [Section 7.1.1](https://www.rfc-editor.org/rfc/rfc6455#section-7.1.1).

So, if the server initiates a websocket close (awaiting `.close` in our case), but the client does not respond with a Close control frame, our current implementation will wait for at least 20 seconds before closing the transport more forcefully. Ive verfied that all major browsers (chrome, firefox, and safari) reply with a close correctly. I frequently use [`websocat`](https://github.com/vi/websocat) to connect to websockets manually for testing purposes. Using this tool, I've verified what i've explained above. This is not a huge deal, as we expect to handle long running connections (hence the use of websockets to begin with) however, without setting a stricter timeout, we are unnecessarily using resources at both the kernel and more crucially the python async event loop levels.

I recommend that we set the `close_timeout` to something like 2.5 seconds. This would result in closing a connection after ~5 seconds if the client has replied with a Close control frame. 

Effected code:

https://github.com/NOAA-OWP/DMOD/blob/6119d6c5c4fb80cc73c15209021453fa9a1f1f17/python/lib/communication/dmod/communication/websocket_interface.py#L178

```python
import logging

logging.basicConfig(
    format="%(asctime)s %(message)s",
    level=logging.INFO,
)
from datetime import datetime
import asyncio

from websockets.server import WebSocketServerProtocol, serve


async def handle(ws: WebSocketServerProtocol):
    print(f"opened: {datetime.utcnow().isoformat()}")
    await ws.send("goodbye")
    await ws.close()
    print(f"closed: {datetime.utcnow().isoformat()}")


async def main():
    # async with serve(handle, "localhost", 8000, close_timeout=0.5):
    async with serve(handle, "localhost", 8000):
        await asyncio.Future()  # run forever


asyncio.run(main())
```

aaraney:
fixes #574 
see #574 for context.


robertbartel:
How sure are we that there's no scenario where:

1. a client could fail to respond properly due to getting stuck in an async wait 
2. the connection being forcibly closed by the service end causes a cascading issue for the client (which may also be/be running in a different service)

aaraney:
> How sure are we that there's no scenario where:
> 
> 1. a client could fail to respond properly due to getting stuck in an async wait
> 2. the connection being forcibly closed by the service end causes a cascading issue for the client (which may also be/be running in a different service)

Good questions:

1. This timeout is unrelated to the `ping_timeout` / `ping_interval`. This is only when the server is trying to close the connection. If the server sends a ping and the client does no respond with a pong in `ping_timeout` seconds, then the connection will be closed at the transport layer anyways. See [RFC 6455 section 5.2.2](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.2) for more information about this.
2. IMO that is something the client is responsible for handling. For example, the server nor the client can guarantee that the network is reliable (per fallacies of distributed computing).

robertbartel (approved):
What I really care about is:  is this something that may trigger subtle problems that make us want to pull our hair out (I guess I've had too much of that with websockets), and that we don't really have to do yet?

I specifically say "problems" and not "bugs," because I'd expect any trouble this triggered would likely be due to holes in our listeners and clients.  Bugs that are there already; they just aren't being exercised.  So, yeah, I agree, e.g., that clients should be responsible for those things.  I just wanted to be thoughtful about whether it's more optimal to wait to enforce that.  

All that said, we are probably ok here.  I am a little worried about this occasionally contributing to something weird happening, but I don't expect it'd be enough for us to be pulling our hair out.

PR #577, opened 2024-04-11
refactor: remove deprecated path parameter

aaraney:
`path` parameter was deprecated in [`websockets` `v10.1`](https://websockets.readthedocs.io/en/stable/project/changelog.html#id20). Remove this from the `WebSocketInterface` `listen` method and from all implementations.

Im happy to version bump, but im not sure it is necessary. Thoughts, @robertbartel?


robertbartel (changes requested):
To be safe, I think we should to update to requiring >= 10.1 in _python/lib/communication/setup.py_.

aaraney:
Great call, @robertbartel!

robertbartel (approved):

PR #604, opened 2024-05-06
style: remove unused variables and replace f-strings w/out placeholders

aaraney:
- Remove unused variables (all exception variables that are unused)
- Replace f-strings w/out placeholders with normal strings


christophertubbs (approved):
Great work and great initiative.

The only thing that raises an eyebrow is the lack of instantiation of a couple exceptions. That behavior wasn't changed and is therefore out of scope of this ticket, but it might need to be reconsidered if we want that to be a reasonable pattern to allow in terms of basic styling. Python is helpful in that it'll instantiate it if not done so by the writer, but it's still odd behavior that might be surprising to other devs. I know _**I**_ was caught off guard by it.

PR #608, opened 2024-05-09
Update dataset manager API for data-domain-impacting operations

robertbartel:
Updating _DatasetManager_ and implementations to require a _DataDomain_ parameter being provided when either adding data to or deleting data from an existing dataset.  Implementing functions to allow for domain subtraction.  Also fixing domain merge function for consistency, so that it only works on one constraint at a time.

Note that this PR depends on #566 and should remain in draft status until that one is merged.


python/lib/core/dmod/core/meta_data.py
```diff
@@ -718,6 +817,36 @@ def is_all_possible_values(self) -> bool:
...
+            A new instance representing the result of subtraction.
+
+        See Also
+        --------
+        can_have_subtracted
+        """
+        if not self.can_have_subtracted(subtrahend):
+            raise ValueError(f"Can't subtract given {subtrahend.__class__.__name__}")
+        if self.is_all_possible_values or subtrahend.is_all_possible_values:
+            raise ValueError(f"Can't subtract given when dealing with restriction having all values in a context")
```
aaraney, line 840:
```suggestion
            raise ValueError("Can't subtract unbound restriction")
```

aaraney (approved):
Looks good! Thanks for the changes, @robertbartel!

PR #610, opened 2024-05-10
Fix bug in dataservice use of obj store settings

robertbartel:
Fixing usage of ServiceSetting object in service's main lifespan() func, as use was of earlier names for properties, which were changed in code review but not updated here; also bumping package version.


robertbartel:
@christophertubbs, since Austin probably won't be able today, when you have 5 minutes, could you give this a quick review?  It's very small, fixing some uses of attributes of  _ServiceSettings_ (that isn't change and can be found [here](https://github.com/robertbartel/DMOD/blob/dd2dc02dcf0bf3eece1ce9a5b77a2d108720ce9e/python/services/dataservice/dmod/dataservice/service_settings.py)) in the data service, which I think we just missed when we modified some the names during that PR.

christophertubbs:
Any way I can reasonably complain about `"object_store_exec_user_name"` having `exec` in it? Does that mean "Object Store Execution User Name"? Also, is that user name used anywhere else? I would've assumed it should be something along the lines of `access_key` if it's going to be used for a `access_key` parameter.

I imagine this pickiness is out of scope though. Let me know if that's the case.

robertbartel:
A fair question, @christophertubbs, but there is a reason behind it (albeit convoluted).

The Python attributes need to match the names of some configured Docker secrets in the _main_ stack so that the _ServiceSettings_ class can load those attributes' values from the secrets. 

The [_main_ stack Docker secrets](https://github.com/NOAA-OWP/DMOD/blob/27c0e33df30d560fb03721e51fefee753c8d2576/docker/main/docker-deploy.yml#L235) - `object_store_exec_user_name`/`object_store_exec_user_passwd` - are named that way to correspond to analogous secrets in the _object_store_ Docker stack, with "object_store_" prefixed to provide context.

The [_object_store_ Docker stack secrets](https://github.com/NOAA-OWP/DMOD/blob/27c0e33df30d560fb03721e51fefee753c8d2576/docker/object_store/docker-single-node.yml#L75) - `exec_user_name`/`exec_user_passwd` - are name as such to differentiate them from `access_key`/`secret_key`.   The latter are the admin credentials for MinIO, and their names are dictated by the MinIO Docker image.  `exec_user_name`/`exec_user_passwd` are for the user credentials that should be used during normal execution of DMOD.

christophertubbs (approved):
lgtm!

But seriously, good fix.

PR #611, opened 2024-05-10
Fix bugs discovered in testing

robertbartel:
- Fixing issue with tests for object store dataset manager, where an extra line in Docker secrets files for object store credentials would previously break functionality
- Improving integration test helper script for _dmod.modeldata_ to wait for object store services to fully start before running the tests that needed them started
- Fixing an import related bug in _dmod.scheduler_ tests that was breaking behavior through the tests, because an `isinstance` call was failing due to the same _actual_ class getting imported different ways in different parts of the code
- Fix a bug in the recently updated round robin allocation function, to make sure it actually collects and returns the allocations that it allocates


aaraney (approved):
LGTM! We should move configuration of this service over to `pydantic.BaseSettings` similar to the `dataservice`.

PR #612, opened 2024-05-13
Fix bind host value setup for data service.

robertbartel:
Add env var `HOST` to _data-service_ Docker config, as the entrypoint script requires this in the environment. 
Having this Docker env var be sourced from a DMOD deployment env var `DOCKER_DATA_SERVICE_LISTEN_HOST` (or fall back to default of 0.0.0.0) and adding this new DMOD env var to example.env.

Also, fixing typo in start of _uvicorn_ in entrypoint script, where there was no space between bind host arg and `--port` flag.


aaraney (approved):
Looks good!

PR #620, opened 2024-05-16
Ensure memory can be set for requested jobs

robertbartel:
Updating request objects and CLI to make sure that `memory` can be explicitly provided when requesting a job.


aaraney:
In taking a quick look at this, I know the 500Kb default memory limit originates elsewhere, but we should really raise that default value. I suggest something like 128, 256, or 512MiB instead or even higher.

robertbartel:
@aaraney, that's a bit of a can of worms.  I've opened #621 for us to think about that and track potential changes.

PR #622, opened 2024-05-17
Fix RealizationConfigDomainDetector file opening

robertbartel:
Fixing bug in opening of file to deserialize it for domain detection.  Also adding unit tests for this class.


christophertubbs (approved):
roll tide on those unit tests.

PR #624, opened 2024-05-17
Fix ngen entrypoint script for geopackage hydrofabrics

robertbartel:
Fixing to properly support geopackage hydrofabric dataset type in commands to execute ngen if that is what is provided.

Also adding creation of a symlink to output dataset in fixed location for routing.

### DIFF
--- a/docker/main/ngen/funcs.sh
+++ b/docker/main/ngen/funcs.sh
@@ -92,15 +92,24 @@ exec_main_worker_ngen_run()
 
     # Execute the model on the linked data
     echo "$(print_date) Executing mpirun command for ngen on ${MPI_NODE_COUNT} workers"
-    ${MPI_RUN:?} -f "${MPI_HOSTS_FILE}" -n ${MPI_NODE_COUNT} \
-        ${NGEN_EXECUTABLE:?} ${HYDROFABRIC_DATASET_DIR}/catchment_data.geojson "" \
-                ${HYDROFABRIC_DATASET_DIR}/nexus_data.geojson "" \
-                ${CONFIG_DATASET_DIR}/realization_config.json \
-                ${PARTITION_DATASET_DIR}/partition_config.json \
-                --subdivided-hydrofabric
-
-    #Capture the return value to use as service exit code
-    NGEN_RETURN=$?
+    if [ -e ${HYDROFABRIC_DATASET_DIR}/catchment_data.geojson ]; then
+        ${MPI_RUN:?} -f "${MPI_HOSTS_FILE}" -n ${MPI_NODE_COUNT} \
+            ${NGEN_EXECUTABLE:?} ${HYDROFABRIC_DATASET_DIR}/catchment_data.geojson "" \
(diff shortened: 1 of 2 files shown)


docker/main/ngen/ngen_entrypoint.sh
```diff
@@ -48,6 +48,10 @@ init_ngen_executable_paths
 
 # Move to the output dataset mounted directory
 cd ${OUTPUT_DATASET_DIR:?Output dataset directory not defined}
+#Needed for routing
```
christophertubbs, line 51: Is this hard coding intentional? I don't see any other references to `/dmod/dataset/experiment_output` and we should probably stay away from just naming stuff some form of `experiment` - we're still stuck with that in the GUI.

docker/main/ngen/funcs.sh
```diff
@@ -92,15 +92,24 @@ exec_main_worker_ngen_run()
...
-    ${MPI_RUN:?} -f "${MPI_HOSTS_FILE}" -n ${MPI_NODE_COUNT} \
-        ${NGEN_EXECUTABLE:?} ${HYDROFABRIC_DATASET_DIR}/catchment_data.geojson "" \
-                ${HYDROFABRIC_DATASET_DIR}/nexus_data.geojson "" \
-                ${CONFIG_DATASET_DIR}/realization_config.json \
-                ${PARTITION_DATASET_DIR}/partition_config.json \
-                --subdivided-hydrofabric
-
-    #Capture the return value to use as service exit code
-    NGEN_RETURN=$?
+    if [ -e ${HYDROFABRIC_DATASET_DIR}/catchment_data.geojson ]; then
```
christophertubbs, line 95: These are probably more evidence that we need to bump the priority for the ticket for moving logic into python. It might just be because I generally avoid too much shell scripting, but I find this hard to read/track. It doesn't break convention with the rest of the script, so moving that out to another language isn't in the scope of this task.

christophertubbs (changes requested):
The symbolic link for routing is questionable in my eyes. If it's important to rubberstamp a path named `experiment_output` and not something that stands to be a little more longer lasting, cool, I just want to make sure it's acknowledged.

docker/main/ngen/ngen_entrypoint.sh
robertbartel: The hard coding is intentional. In short, we need to configure something in advance that will line up with something not known until runtime. 

A troute config file needs to include the path to the ngen output (and its own output).  But job worker have datasets mounted in as a directory with the same name as the dataset itself. Output dataset names are based on the job id, and that is dynamically generated (at least usually). 

I realized after setting this up that we don't absolutely have to do this now, but I decided to leave it in. I could take it out, but it's a problem we eventually will have to deal with, in which case, why not at least do this for now?

I also don't object to changing the name of the symlink, if you have a strong preference about that.

christophertubbs: Hey, there's a ticket, so I'm good.

PR #629, opened 2024-05-21
Workflow fixes

robertbartel:
Several miscellaneous fixes found while testing the DMOD model exec workflow and other related supplementary CLI commands.


docker/main/ngen/Dockerfile
```diff
@@ -664,6 +664,10 @@ RUN if [ "${NGEN_WITH_PYTHON}" == "ON" ]; then \
     fi
 USER ${USER}
 
+ENV VIRTUAL_ENV=/dmod/venv
+RUN python3 -m venv $VIRTUAL_ENV && pip3 install numpy
```
aaraney, line 668: Is it intentional not to activate the `venv` after creating it here? So, should this instead be:
```suggestion
RUN python3 -m venv $VIRTUAL_ENV && source $VIRTUAL_ENV/bin/activate && pip3 install numpy
```

docker/main/ngen/Dockerfile
```diff
@@ -664,6 +664,10 @@ RUN if [ "${NGEN_WITH_PYTHON}" == "ON" ]; then \
...
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
```
aaraney, line 669: Hmmm, I assume you are adding this to `PATH` so `venv` is effectively always activated. Is there a reason why we can't use the default python environment _or_ instead add `venv` activation in the entrypoint?

docker/main/ngen/Dockerfile
```diff
@@ -863,6 +871,13 @@ ENV WORKDIR=${WORKDIR}
 ENV HYDRA_HOST_FILE=/etc/opt/hosts
 ENV PATH=${WORKDIR}:${WORKDIR}/bin:/dmod/bin:${PATH}:/usr/lib64/mpich/bin
 
+ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/lib:/usr/local/lib64:/dmod/shared_libs
+# This next value is eventually needed for the t-route build ...
+# ... make sure it stays in sync with configure step for netcdf above
+ENV NETCDFINC=/usr/include
+ENV VIRTUAL_ENV=/dmod/venv
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
```
aaraney, line 879: similar question as above here.

aaraney (commented):
Had one question that should be super easy to address. Lets get this on the road though. Thanks for making these changes @robertbartel!

docker/main/ngen/Dockerfile
robertbartel: Pretty sure it is effectively active if `VIRTUAL_ENV` is set in the environment.

docker/main/ngen/Dockerfile
robertbartel: Basically, something weird seems to be going on, probably after #755, that's breaking the image build.  See https://github.com/CIROH-UA/NGIAB-HPCInfra/issues/12 and https://github.com/CIROH-UA/NGIAB-CloudInfra/issues/137 for more details on this specifically, as others are running into it also.

docker/main/ngen/Dockerfile
aaraney: I think that is just used for bookkeeping so that the added `deactivate` shell function can remove `$VIRTUAL_ENV` from `$PATH`.

docker/main/ngen/Dockerfile
```diff
@@ -664,6 +664,10 @@ RUN if [ "${NGEN_WITH_PYTHON}" == "ON" ]; then \
     fi
 USER ${USER}
 
+ENV VIRTUAL_ENV=/dmod/venv
+RUN python3 -m venv $VIRTUAL_ENV && pip3 install numpy
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
 RUN cd ${BOOST_ROOT} \
```
aaraney, line 671: It seems like activating the virtual environment here before doing the build and then change 856 to `ENTRYPOINT ["source $VIRTUAL_ENV/bin/activate", "&&", "entrypoint.sh"]`?

docker/main/ngen/Dockerfile
aaraney: Thanks for linking those related issues. Seems like something to do with using a non-virtual environment. My main concern is making it really clear that we are using a virtual environment in the image without needing to come and read the `Dockerfile`.

robertbartel: Fair point.  I've opened #630.  I don't really want to move away from the default environment, but it seems necessary.  But I'd like to eventually take this out, once the underlying ngen build issues are resolved.

docker/main/ngen/Dockerfile
robertbartel: Can you clarify what you are asking here?

docker/main/ngen/Dockerfile
aaraney: Yeah, this is super weird. Lets just move ahead like you are suggesting and revisit this when we have a fix.

aaraney (approved):

PR #634, opened 2024-05-28
Fix resource config script for missing directory

### LINKED ISSUE #626: If the expanded `SCHEDULER_RESOURCE_DIR` directory does not exist, `scripts/create_resources_config.sh` `:60` will fail
https://github.com/NOAA-OWP/DMOD/blob/417ea09dcd2cfedc8ea1225389d170801dc75c49/scripts/create_resources_config.sh#L60

Should we warn and create this directory, @robertbartel?

Aside, for context, im following the [setup single-node deployment guide](https://github.com/NOAA-OWP/DMOD/wiki/Setting-Up-a-Single%E2%80%90Node-DMOD-Deployment). Using [`scripts/create_env_config.sh`](https://github.com/NOAA-OWP/DMOD/blob/417ea09dcd2cfedc8ea1225389d170801dc75c49/scripts/create_env_config.sh#L52) to create a `.env` file as noted in the guide, you will get `SCHEDULER_RESOURCE_DIR =${REPO_ROOT_DIR}/data/scheduler_service` in your `.env` file. I don't see anywhere where this should be created in previous steps. This is what leads me to think we should either just create the directory if it is missing here, or create the directory when the `.env` file is created. Thoughts?

robertbartel:
Updating create_resources_config.sh helper script to account for missing parent directory for the to-be-created resource.yaml and create that directory when needed, or bail if the path is an existing non-directory.

Closes #626.

### DIFF
--- a/scripts/create_resources_config.sh
+++ b/scripts/create_resources_config.sh
@@ -52,7 +52,12 @@ while [ ${#} -gt 0 ]; do
     shift
 done
 
-if [ -e ${SCHEDULER_RESOURCE_DIR}/resources.yaml ]; then
+if [ ! -e ${SCHEDULER_RESOURCE_DIR:?Scheduler resources directory not configured in environment} ]; then
+    mkdir -p ${SCHEDULER_RESOURCE_DIR}
+elif [ ! -d ${SCHEDULER_RESOURCE_DIR} ]; then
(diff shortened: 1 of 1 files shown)


aaraney (approved):

Issue #637, opened 2024-05-29
Investigate IO performance issue with workers and object-store-backed datasets


robertbartel:
In recent testing, analogous jobs took several orders of magnitude longer when workers used object-store-backed datasets compared to if they used data supplied via local (manually synced) bind mounts.  Rough numbers for one example are 5 hours compared to 5 minutes.

These tests involved a single node with 24 processors and 96 GB of memory, running both the services and workers.  I.e., it's entirely possible the performance deficiency is heavily related to the compute infrastructure being insufficient for running jobs and an object store like this (initial development of these features involved 2-5 data-center-class servers).  On the other hand, this is not an unrepresentative use-case for what we want DMOD to do.

Some investigation is needed as to whether there are any viable means to at least reasonably mitigate these issues (i.e., longer runtimes on lesser hardware are to be expected, but going from 5 minutes to 4 hours just isn't acceptable).   Unless these performance issues are drastically reduced, we should likely accelerate plans for developing alternative dataset backends, such as cluster volumes within the Docker Swarm (see #593).

christophertubbs:
Caching may help, but memory configuration and VM vs container vs bare metal may make a difference. Totally different products, but CEPH had issues if it had too _much_ RAM. It had something to do with data replication. Don't remember - it was a nightmare. Reading and writing from the object store may also cause problems due to rapid io on the part of ngen. If there's just a little bit of extra overhead on the object store calls and ngen is opening and closing a bajillion files, it might be crippling. If it comes down to it, a pre and post processing step to stage the required data locally might be a solution. That'll be its own nightmare.

robertbartel:
> Reading and writing from the object store may also cause problems due to rapid io on the part of ngen. If there's just a little bit of extra overhead on the object store calls and ngen is opening and closing a bajillion files, it might be crippling.

This is basically my initial guess for the problem:  too much IO, applied to only a single MinIO node now (due to other restrictions; see #242 et al.), running on older, less robust hardware.  It's not out of the question that the object store proxy or Docker storage plugin configs could be sub-optimal, or that something else is going on, but I'm not getting my hopes up.

>  If it comes down to it, a pre and post processing step to stage the required data locally might be a solution. That'll be its own nightmare.

[...]

PR #650, opened 2024-06-06
Optimize worker behavior for object store writes

robertbartel:
Optimizing workers for object store writes (i.e., the only current dataset backing type), essentially by having them archive large collections of files to be written - e.g. output CSVs - and only write these archives - i.e., minimal individuals files to the object store.


robertbartel:
Relates to #637.

python/lib/core/dmod/core/meta_data.py
```diff
@@ -221,15 +221,28 @@ class DataFormat(PydanticEnum):
...
+                           StandardDatasetIndex.DATA_ID: None},
+                          None,
+                          True)
+    """ Format for output of ngen when written to single NetCDF file, with dynamically configured output fields. """
+
     @classmethod
     def can_format_fulfill(cls, needed: DataFormat, alternate: DataFormat) -> bool:
         """
-        Test whether data in an alternate format is capable of satisfying requirements of some other format.
+        Test whether a dataset and contained data in some format can satisfy requirements of a different format.
```
christophertubbs, line 240: Props to the attention to documentation

python/services/dataservice/dmod/dataservice/service.py
```diff
@@ -787,11 +787,18 @@ def _create_output_datasets(self, job: Job):
                     break
 
             # TODO: (later) more intelligently determine type
-            mgr = self._managers.manager(DatasetType.OBJECT_STORE)
+            dataset_type = DatasetType.OBJECT_STORE
```
christophertubbs, line 790: This is waaaay better than it was. Good job!

christophertubbs (changes requested):
tar_and_copy and its related functions needs to be separated out into its own python script. There's not much keeping the commands `tar_and_copy --dry-run --compress ${CONFIG_DATASET_DIR:?Config dataset directory not defined} config_dataset.tgz ${OUTPUT_DATASET_DIR:?}` and `./tar_and_copy.py --dry-run --compress ${CONFIG_DATASET_DIR:?Config dataset directory not defined} config_dataset.tgz ${OUTPUT_DATASET_DIR:?}' from yielding the same results.

docker/main/ngen/funcs.sh
```diff
@@ -212,4 +212,95 @@ run_secondary_mpi_ssh_worker_node()
...
+        >&2 echo "$(print_date) Error: tar_and_copy contents directory '${_CONTENTS_DIR}' exists but is not a directory!"
+        exit 1
+    elif [ ! -e "${_DEST_DIR}" ]; then
+        >&2 echo "$(print_date) Error: tar_and_copy copy destination directory '${_DEST_DIR}' does not exist!"
+        exit 1
+    elif [ ! -e "${_DEST_DIR}" ]; then
+        >&2 echo "$(print_date) Error: tar_and_copy copy destination directory '${_DEST_DIR}' exist but is not a directory!"
+        exit 1
+    elif [ -e "${_TAR_FILE}" ]; then
+        >&2 echo "$(print_date) Error: tar_and_copy archive file '${_TAR_FILE}' already exists!"
```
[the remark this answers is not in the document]
robertbartel: In a vacuum, yes, but this is specifically intended to be run inside a fresh container. So there  really should never be anything there already. Chances are that if there is, something is going wrong.

docker/main/ngen/funcs.sh
```diff
@@ -212,4 +212,95 @@ run_secondary_mpi_ssh_worker_node()
     while [ -e ${RUN_SENTINEL} ] && kill -s 0 "${_SSH_D_PID}" 2>/dev/null ; do
         sleep 5
     done
+}
+
+tar_and_copy()
```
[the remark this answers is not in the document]
robertbartel: So my initial thought was that it didn't make sense to try to do new things as separate pieces in Python. That it probably is better overall for these things to be done in Python, rather than shell scripts, but that really all the shell scripts (or at least groups that run together in any particular setting, like those involved in container entrypoints) should be ported all at once. 

I'll rethink this a bit. I don't want to introduce something that's on the whole even more complicated to follow than this, and that was my worry. But maybe it won't be quite that bad.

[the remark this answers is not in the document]
christophertubbs: This might do exactly what you want: https://docs.python.org/3.8/library/shutil.html#shutil.make_archive

Converting the scripts to a bit at a time would make things easier. That way we can ensure that each bit works with as small a scope as possible, it'll be less of a burden, it won't take as long, and we'll be less likely to run into errors.

docker/main/ngen/funcs.sh
```diff
@@ -212,4 +212,95 @@ run_secondary_mpi_ssh_worker_node()
...
+        >&2 echo "$(print_date) Error: cannot move output from non-directory path '${1}' to output dataset!"
+        exit 1
+    elif [ ! -d ${2:?No output dataset directory given for copying} ]; then
+        >&2 echo "$(print_date) Error: cannot move output to non-directory path '${1}' for output dataset!"
+        exit 1
+    fi
+
+    if [ $(ls ${1} | grep '.csv' | wc -l) -gt 0 ]; then
+        echo "$(print_date) Archiving and copying output CSVs to output dataset"
+        tar_and_copy ${1} job-${JOB_ID:?}-output.tar ${2}
```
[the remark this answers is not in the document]
robertbartel: I think because the goal was to reduce practical running time by reducing transferred file count specifically.  The size of the data was not of tremendous importance.  Compression would slow things, both now and later.  In contrast, since the configs were effectively duplicated here, and not the primary content of the output dataset, it made more sense to compress that data.

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
```diff
@@ -311,7 +311,7 @@ def output_formats(self) -> List[DataFormat]:
         List[DataFormat]
             List of the formats of each required output dataset for the requested job.
         """
-        return [DataFormat.NGEN_OUTPUT]
+        return [DataFormat.ARCHIVED_NGEN_CSV_OUTPUT]
```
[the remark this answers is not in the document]
robertbartel: An interesting question.  Probably something that deserves its own issue, as it isn't well defined at the moment.

docker/main/ngen/py_funcs.py
```diff
@@ -0,0 +1,309 @@
...
+    sub_cmd_parser = parent_subparser_container.add_parser('to_directory', description="Move to a specified directory")
+    sub_cmd_parser.add_argument("dest_dir", type=Path, help="Destination directory to which to move the output")
+
+
+def _parse_for_move_job_output(parent_subparsers_container):
+    # A parser for the 'tar_and_copy' param itself, underneath the parent 'command' subparsers container
+    desc = "Move output data files produced by a job to another location, typically to put them into a DMOD dataset."
+    helper_cmd_parser = parent_subparsers_container.add_parser('move_job_output', description=desc)
+    helper_cmd_parser.add_argument('--job_id', '--job-id', dest='job_id', help='Optionally specify job id.')
+    helper_cmd_parser.add_argument('--archive-files', dest='do_archiving', choices=["true", "false"],
```
christophertubbs, line 57: Set boolean values via `action="store_true"`, not via choices and a function in the type.

christophertubbs: on the other hand, if this is supposed to be called programmatically via a bash function, set up a script to call this programmatically via python, _not_ by passing raw boolean values through the CLI.

docker/main/ngen/py_funcs.py
```diff
@@ -0,0 +1,309 @@
...
+def _parse_args() -> argparse.Namespace:
+    """
+    Set up and run top-level arg parsing for module.
+
+    Returns
+    -------
+    argparse.Namespace
+        The parsed arguments namespace object.
+    """
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter, prog='py_funcs')
```
christophertubbs, line 77: Make sure to include a description.

christophertubbs (changes requested):
I'm still exploring some functionality that I'm not sure about with the python ArgumentParser. I may have further comments later

python/lib/communication/dmod/communication/maas_request/ngen/abstract_nextgen_request.py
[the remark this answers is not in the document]
robertbartel: This may be simpler than I thought initially (although we can still consider giving it its own issue if more conversation is needed).

First, it is not possible to completely, explicitly define an ngen output data format that covers all possible output variables.  I'm not going to work through it entirely right now, but my back-of-the-envelop math says ngen may actually support uncountably infinitely many possible combinations of outputs, in theory at least.  Regardless, too many to cover.  

And so, we don't try to.  While some _DataFormat_ values do define `data_fields`, the ngen output formats do not.  The implication is that we cannot necessarily completely summarize what could be in there, so users of the data need to determine that for themselves.

A format does have to enable domains to elaborate on certain things, though, via the standard indices.  So it is useful (perhaps necessary) to distinguish particulars of how the indices data are available within the dataset, and this is done implicitly by the format and its name.  E.g., for `ARCHIVED_NGEN_CSV_OUTPUT`, there will be a bunch of CSV inside an archive, with catchment ids in the names of the CSVs and times (and other data) inside the CSVs. 

So, it is still useful to have ngen-related output formats that imply only part of how data are structured within such datasets, even if it doesn't (because it can't) tell us everything about what's in there.

docker/main/ngen/py_funcs.py
robertbartel: So the reason this was done like this is to have a ``True``: _force-on_, ``False``:_force-off_, ``None``:_dynamically-determine_ setup.  

I wanted this to be able to receive ``None`` downstream where the ``do_archiving`` arg value gets used, as that's when there's some code used to decide whether or not to archive based on the files themselves.   But I also wanted an explicit indication, either way, to be respected.  And using ``store_true`` can't get us both ``None`` and ``False`` values.

Whether or not this is a good idea is certainly open for discussion, but that was why this argument wasn't set up in a more typical manner.

docker/main/ngen/py_funcs.py
```diff
@@ -0,0 +1,309 @@
...
+    parser.add_argument('--log-level', '-L', dest='log_level',
+                        default=os.environ.get("DEFAULT_LOG_LEVEL", "INFO").upper(),
+                        choices=["CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"], help='Optionally specify log level.')
+
+    subparsers = parser.add_subparsers(dest='command', help='Specify the Python helper function command to run.')
```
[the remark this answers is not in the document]
christophertubbs: how exclusive are these three operations supposed to be? Is this a replacement for `tar_and_copy` or a replacement for `tar_and_copy` plus other stuff?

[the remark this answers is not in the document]
robertbartel: Initially I'd added a handful of new shell functions within _funcs.sh_, one of which was `tar_and_copy`.    I didn't want to change the entire design of the entrypoint setup too much at this stage, so I added _py_funcs.py_ and rewrote all of those new functions in Python there (removing the original additions to _funcs.sh_).  There is, of course, more in _py_funcs.py_, mostly for CLI arg parsing this file can be used directly from the existing shell scripts, but also a little bit to modularize some of the behavior.

robertbartel:
The tests that are failing seem to be in the data service tests and related (somehow) to the recent move to REST:

> ModuleNotFoundError: No module named 'fastapi'

I find this peculiar, since there have been many PRs since #575, so surely that module was present during  Github Action checks for the PRs between that and this.  `fastapi` is included in `install_requires` within _python/services/dataservice/setup.py_, though it isn't in _requirements.txt_.

I'd feel like we could probably disregard it here.  @aaraney and @christophertubbs, any thoughts on either the disregarding or root cause here?

christophertubbs:
It might be worth putting a reference to fastapi in the requirements as a cya manuever, but I'm not sure if that's a big enough reason to be a huge blocker as long as an "Urgent"/"Low Hanging Fruit" ticket is added for it.

Did that cause the other issues where programs didn't know what `dmod` was?

docker/main/ngen/py_funcs.py
christophertubbs: I _think_ I know a good way of doing this, but I need to poke around at it first.

docker/main/ngen/py_funcs.py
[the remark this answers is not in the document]
christophertubbs: I'm _really_ not a fan of shoving the kitchen sink under here (rather than having `tar_and_copy.py` for instance), but if this is a case of "Perfection is the enemy of progress", I'm all for slapping a bunch of warnings around it and adding a ticket saying something along the lines of "spruce this up this way".

docker/main/ngen/py_funcs.py
christophertubbs: Code like the following should work _and_ provide a more elegant cli:

```python
    archive_group = helper_cmd_parser.add_mutually_exclusive_group()
    archive_group.add_argument(
        "--force-archives",
        dest="force_archives",
        action="store_true",
        help="Whatever the force on logic is"
    )
    archive_group.add_argument(
        "--prevent-archives",
        dest="prevent_archives",
        action="store_true",
        help="Whatever the force off logic is"
    )
```

You'd then modify:

```python
def move_job_output(output_directory: Path, move_action: str, do_archiving: Optional[bool] = None,
                    job_id: Optional[str] = None, **kwargs):
```

to be

```python
def move_job_output(output_directory: Path, move_action: str, force_archiving: bool = False,
                    prevent_archives: bool = False, job_id: Optional[str] = None, **kwargs):
```

which will have logic like:

```python
    # If this was not set, dynamically determine what it should be
    if not force_archiving and not prevent_archives:
        # For now, just do this if the output data contains more than 100 individual files
        out_dir_files = [f for f in output_directory.glob("**/*")]
        out_dir_file_count = len(out_dir_files)
```

If you want to make things a tad more explicit, you can use an enum:

```python
class ArchiveStrategy(enum.Enum):
    FORCE_ON = "force"
    FORCE_OFF = "prevent"
    DYNAMIC = "dynamic"

...

    helper_cmd_parser.add_argument(
        "--archiving",
        dest="archiving",
        default=ArchiveStrategy.DYNAMIC,
        type=ArchiveStrategy,
        help="Limited value with an enum"
    )

...

def move_job_output(output_directory: Path, move_action: str, archiving: ArchiveStrategy = ArchiveStrategy.DYNAMIC,
    job_id: Optional[str] = None, **kwargs):

...
   if archiving == ArchivingStrategy.DYNAMIC:
        # For now, just do this if the output data contains more than 100 individual files
        out_dir_files = [f for f in output_directory.glob("**/*")]

```

docker/main/ngen/py_funcs.py
```diff
@@ -0,0 +1,310 @@
...
+    Get the current date and time as a string with format ``%Y-%m-%d,%H:%M:%S``
+
+    Returns
+    -------
+    The current date and time as a string.
+    """
+    return datetime.now().strftime('%Y-%m-%d,%H:%M:%S')
+
+
+def move_job_output(output_directory: Path, move_action: str, do_archiving: Optional[bool] = None,
```
christophertubbs, line 158: `Optional[T] = None` is redundant. `do_archiving: Optional[bool] = None` is the same as `do_archiving: bool = None` is the 95% the same thing (that other 5% being when no default value is given - then it's just `Union[bool, None]' and requires the value).

Dropping the `Optional` annotation results in clearer code

docker/main/ngen/py_funcs.py
```diff
@@ -0,0 +1,310 @@
...
+
+    scp_subs = OrderedDict()
+
+    for host in (h for h in mpi_host_names if h != local_hostname):
+        scp_subs[host] = Popen(f"scp -q -r {host}:${output_write_dir!s}/ ${output_write_dir!s}/.")
+
+    for host in scp_subs:
+        return_code = scp_subs[host].wait()
+        if return_code != 0:
+            raise RuntimeError(f"{get_date_str()} gather_output failed for {host} w/ return code {return_code!s}!")
```
christophertubbs, line 144: Try the following instead:

```python
for host, process in scp_subs.items():
    _, error_in_bytes = process.communicate()
    if process.returncode != 0:
        raise RuntimeError(f"{get_date_str()} gather_output failed for '{host}': {error_in_bytes.decode()} (code={process.returncode}!")
```

docker/main/ngen/py_funcs.py
```diff
@@ -0,0 +1,310 @@
...
+    mpi_host_names
+        List of MPI host names for the relevant job.
+    output_write_dir
+        Common job output directory across all hosts, from which data on remotes should be taken and in which data on
+        this host should be collected.
+    """
+    from socket import gethostname
+    local_hostname = gethostname()
+
+    scp_subs = OrderedDict()
```
christophertubbs, line 136: Why is an `OrderedDict` used here? as of 3.7, `dict` and `OrderedDict` are nearly identical. The only major differences are how `OrderedDict`s are considered equal to other `OrderedDict`s (but not other maps), a few extra operators used for merging, and different behavior for `popitem` and `reversed`.

christophertubbs (changes requested):

docker/main/ngen/py_funcs.py
aaraney: I disagree. Without `Option[T]` a linter will yell at you for checking if it `is None`. I think the semantics of using an optional boolean as a ternary operation _is_ confusing though. Or at least, im going to look at it twice to realize what's going on. Probably clearest to make this into an `enum` and set the default keyword argument to an enum member.

docker/main/ngen/py_funcs.py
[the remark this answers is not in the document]
robertbartel: Yeah, this goes back to my comment earlier about #665 and why I originally wrote things in bash.  There is no way to make this look good (at least not to all of us at once) without going waaay out of scope.  But we'll get there, and I think this approach will make that a least a little easier (given that the changes had to be done somehow).

docker/main/ngen/py_funcs.py
robertbartel: I don't actually think this is better here.  For one, I don't particularly like that this leads to the possibility, and thus the need to handle the condition, when:

`force_archiving == True and prevent_archives == True`

I'd rather structure things to avoid the possibility of the contradiction altogether.

Beyond that, we also can assume we'll be changing how the CLI parsing and behavior works when we do #665.  Aside from this upcoming flux complicating a plan to prevent the aforementioned contradiction at the CLI level, I'd rather prioritize the design of the underlying function in this case.

I'm also a little on the fence as to whether `--force-this` / `--disallow-this` is generally easier to use and maintain than `--do-this [true|false]`.  I could certainly be convinced there, but I think that's another thing that should wait until #665.

docker/main/ngen/py_funcs.py
```diff
@@ -0,0 +1,310 @@
...
+
+    if args.command == 'tar_and_copy':
+        tar_and_copy(**(vars(args)))
+    elif args.command == 'gather_output':
+        mpi_host_to_nproc_map = process_mpi_hosts_string(args.mpi_hosts_str)
+        gather_output(mpi_host_names=[h for h in mpi_host_to_nproc_map], output_write_dir=args.output_write_dir)
+    elif args.command == 'move_job_output':
+        move_job_output(**(vars(args)))
+    else:
+        raise ValueError("Unsupported command {}".format(args.command))
```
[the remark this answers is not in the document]
robertbartel: Strictly speaking, yes, but I like this for future proofing.  I'm one to occasionally forget the "connecting wiring" between things, like adding the lines in `main` between the arg parsing for a new function and the function implementation itself.

Would it be more appropriate if I changed this to `NotImplementedError`?

docker/main/ngen/py_funcs.py
robertbartel: I made a slight tweak to the error message for line formatting reasons but essentially adopted your suggestion.

docker/main/ngen/py_funcs.py
robertbartel: Gotcha.  I've taken it out.

docker/main/ngen/py_funcs.py
robertbartel: Related to an earlier comment, I think the `Optional[bool]` ternary is preferable to a `force_this: bool, disallow_this: bool`  alternative due to the imprecision and potential contradiction of the latter.  The enum approach seems alright, except that it's causing the scope to keep growing on something intended to be a stopgap because it was already out of scope.  That said, if both of you think the ternary approach is too undesirable (even in the short term), let me know and I can change things to be enum based.

docker/main/ngen/py_funcs.py
christophertubbs: I lurch away from `--do-this [true|false]` because it's a soft enum and/or treating a boolean as a qbit. It's not incorrect or non-functional per se, but it's still something that I'd consider bad practice. If you're going to deal with boolean represented values in a CLI, that's generally relegated to the world of flags and a set of constants would generally be a set of constants even when the three states of a boolean may be used to represent the constant.

For your bit about `force_archiving == True and prevent_archives == True` - the mutually exclusive group will prevent that case from ever occurring - the `ArgumentParser` will throw an error before you ever get there.

I'll leave it to Austin to decide if I'm being a little too hardlined on this detail. I don't want an opinion about an interface philosophy to derail anyone's work.

docker/main/ngen/py_funcs.py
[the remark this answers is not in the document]
christophertubbs: Yeah - `NotImplementedError` is more appropriate here. Adding a whole sub parser and not handling it in `main` is a big no-no and should be caught in PR, but this is perhaps the only way to catch a mess up of that magnitude.

That scenario is also why subparsers should be avoided, though.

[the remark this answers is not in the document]
robertbartel: Eh, actually, I went and read the documentation for _NotImplementedError_ after suggesting it.  I'm not sure it's more appropriate here:

> In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.

Aside from outright forgetting, raising some kind of exception here also protects against typos in the compared-to literal strings in the above expressions; otherwise, things would just silently not run (but also not actually fail).  I've switched this to a _RuntimeError_ and modified the message a bit.

docker/main/ngen/py_funcs.py
robertbartel: FYI, I went ahead and put in the enum solution here.

christophertubbs (approved):

christophertubbs:
Looks solid to me. I removed my flag for requested changes. If Austin is good, I'm good.

Issue #654, opened 2024-06-11
Test and optimize IO performance when automatically generating BMI init configs


robertbartel:
While integration of BMI init config auto-generation capabilities was done in #607, practical performance testing was not conducted.  Given #637 and the fact that DMOD currently only implements an object store dataset backing, there may be some practical issues with the current implementation; e.g., it produces configs perfectly correctly, but takes an impractical or excessive (compare to the job needing the configs) amount of time to complete.

First, analysis is needed for the running time in various scenarios, given the current implementation and more practical off-the-shelf hardware configuration (i.e., at most, a small cluster of desktop-level machines).  Depending on the results, adjustments to the implementation should be made to optimize it for current dataset capabilities.  Where possible, this should be done in a way that lends itself well to future dataset backings (i.e., #593), which may or may not have the same IO performance characteristics and thus may need (or benefit from) certain differences in the implementation.

aaraney:
I added support for writing config files to various archive formats in [this](https://github.com/NOAA-OWP/ngen-cal/pull/115) PR. [Here](https://github.com/NOAA-OWP/ngen-cal/blob/master/python/ngen_config_gen/examples/write_to_tar_gz/main.py) is an example of writing config files on the fly to a gzipped archive file. Compression is not required (and of course slows things down). I would be interested to see the performance of writing to just a `tar` archive. 

[This](https://github.com/NOAA-OWP/ngen-cal/issues/117) issue will also be useful in conducting benchmarks. The issue shows an, albeit naive, approach to generating config files concurrently.

PR #671, opened 2024-07-03
Optimize for better IO performance during BMI init config dataset generation

robertbartel:
Making some adjustments to the way the integration with _ngen-cal_'s BMI init config generation capabilities are used to create `BMI_CONFIG` DMOD datasets.  Primary goal was to reduce the amount of time for creating these, especially during job workflow execution (e.g., during an ngen job).  

Previous implementation of on-the-fly BMI dataset generation created and wrote files one at a time, which for object-store-backed dataset took too long (about 1 hour for VPU01 catchments creating Noah-OWP-Modular and CFE configs for each catchment).   Updated implementation writes all configs first, adds them to the dataset all at once, and utilizes some new optimization within the object store dataset manager.  The same operation with the new implementation takes a little less than 1 minute.

Relates to #654.

## Additions

- New _StandardDatasetIndex_ enum value `HYDROFABRIC_DATA_ID` for referencing associated hydrofabric dataset from derived datasets, in particular `BMI_CONFIG` datasets
- New _DataArchiving_ enum type to define supported file archiving methods, in particular within datasets
- New mostly comply integration test _IntegrationTestDataDeriveUtil_ for BMI init config generation capabilities of _DataDeriveUtil_ when used with object store dataset backing
  - Skipped unless explicitly activated via `.test_env` file
  - Requires hydrofabric with attribute data; need reasonable size subset complete with attribute data (future)
- Adding another realization config for use with aforementioned integration test
  - Used to create dataset within object store manager backend

## Changes

- Update to `BMI_CONFIG` _DataFormat_ with indices to reference realization config and hydrofabric datasets, in particular when such datasets are used for BMI init config derivation/generation
- Add optional _DataArchiving_ attribute (and some associated functions) to _Dataset_ to tracking when all the contents of a dataset are wrapped within an archive
- Optimize _BmiAutoGenerationAdder_ to write files first then add to dataset all at once, via _DatasetManager_ for dataset, to give the manager a chance to perform any implementation-specific optimizations
  - Also makes generated dataset read-only (required for manager optimizations discussed below) 
- Optimize _ObjectStoreDatasetManager_'s `add_data` function implementation so that in certain conditions it optimizes by writing all files to an archive and storing this archive file inside the dataset
  - Takes advantage of this capability of MinIO: https://blog.min.io/small-file-archives/
  - Dataset must be set as read-only
  - Dataset must be empty
  - Data supplied must be provided as a directory containing one or more files
- Improve/modularize BMI config generation functionality within _DataDeriveUtil_ 
- Update _InitialDataAdder_ implementations to simply apply the original domain as the "update domain" portion when calling _DatasetManager_.`add_data()`
  - A merge of equal domains will not keep the domain the same 

## Testing

1. Manual testing of functionality via _IntegrationTestDataDeriveUtil_ and manually-set-up VPU01 hydrofabric dataset, with acceptable generation times (under 60 seconds)

## Notes

-

## Todos

- Get updated hydrofabric and finish test class
- Update workers to account for possibility of archived or non-archived dataset

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows project standards (link if applicable)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:


python/lib/modeldata/dmod/modeldata/data/object_store_manager.py
```diff
@@ -292,8 +302,22 @@ def add_data(self, dataset_name: str, dest: str, domain: DataDomain, data: Optio
             msg = "{}.{} source path '{}' does not exist."
             raise ValueError(msg.format(self.__class__.__name__, _getframe(0).f_code.co_name, source))
         elif src_path.is_dir():
-            bucket_root = kwargs.get('bucket_root', src_path)
-            self._push_files(bucket_name=dataset_name, dir_path=src_path, bucket_root=bucket_root)
+            # Recognized scenario when we may want to take advantage of small file archives
+            # (see https://blog.min.io/small-file-archives/)
+            # Also, we already know from above that, if read-only, dataset must also be empty
+            # TODO: (later) consider whether there is some minimum file count (perhaps configurable) to also consider
+            if self.datasets[dataset_name].is_read_only:
```
aaraney, line 309: This feels like we are injecting a use case specific feature into a general api. For instance, in creating a forcing dataset I would likely mark it as read-only as it is more or less static. I don't think, in that case, it is desirable to zip up a directory of, lets say, netcdf forcing files. Zipping up a directory seems like it should be handled by the specific application and not this api. We will have to rethink how the `data_archiving` attribute is set on a `Dataset` if that is the case though.

One thought is to create an algebraic datatype like a `DataFormat.ARCHIVE[ArchiveFormat, DataFormat]` that is, itself, a `DataFormat` but also wraps an `ArchiveFormat` and a `DataFormat`. It might be a little cumbersome to capture this using a python `Enum`, but we can certainly figure out how to best express the idea in the type system. This could be used to set if a `Dataset` is `data_archiving` or not if it still makes sense to keep that as a top level attribute.

This will move complexity to other places, but I think retain some desirable traits of the system more generally. Im sure there is a simpler way to capture this idea that what ive suggested. I think we should strive for simplicity until we can't.

python/services/dataservice/dmod/dataservice/initial_data_adder_impl.py
```diff
@@ -64,7 +69,8 @@ def _add(path: Path, dest: Optional[str] = None, add_root: Optional[Path] = None
             elif not dest:
                 dest = path.relative_to(add_root)
 
-            if not self._dataset_manager.add_data(dataset_name=self._dataset_name, dest=dest, data=path.read_bytes()):
+            if not self._dataset_manager.add_data(dataset_name=self._dataset_name, dest=dest, data=path.read_bytes(),
```
aaraney, line 72: This seems like a great place to use the reader interface.
```suggestion
            with path.open("rb") as fp:
                if not self._dataset_manager.add_data(dataset_name=self._dataset_name, dest=dest, data=fp,
```

python/services/dataservice/dmod/dataservice/initial_data_adder_impl.py
```diff
@@ -223,13 +231,15 @@ def add_initial_data(self):
...
+        original_domain = self._dataset_manager.datasets[self._dataset_name].data_domain
+
         if self.partial_realization_config is not None:
             raise DmodRuntimeError(f"{self.__class__.__name__} can't have 'None' for partial realization property")
 
         try:
             real_config: NgenRealization = self.build_realization_config_from_partial()
             if not self._dataset_manager.add_data(dataset_name=self._dataset_name, dest=self._REAL_CONFIG_FILE_NAME,
-                                                  data=json.dumps(real_config.json()).encode()):
+                                                  data=json.dumps(real_config.json()).encode(), domain=original_domain):
```
aaraney, line 242: I think this should be, no?

```suggestion
                                                  data=real_config.json().encode(), domain=original_domain):
```

```python
a = real_config.json().encode()
b = json.dumps(real_config.json()).encode()

assert type(json.loads(a)) == dict
assert type(json.loads(b)) == str
```

python/services/dataservice/dmod/dataservice/initial_data_adder_impl.py
```diff
@@ -347,7 +359,7 @@ def _add_realization_config_data(self):
             # In this case, we need to derive a dataset from formulations
             real_config: NgenRealization = self.build_realization_config_from_partial()
             if not self._dataset_manager.add_data(dataset_name=self._dataset_name, dest=real_cfg_file_name,
-                                                  data=json.dumps(real_config.json()).encode()):
+                                                  data=json.dumps(real_config.json()).encode(), domain=original_domain):
```
aaraney, line 362: Same here.

python/services/dataservice/dmod/test/it_data_derive_util.py
```diff
@@ -0,0 +1,199 @@
...
+    explicit_on = os.environ.get("DERIVE_UTIL_IT_ON", "false").strip().lower() == "true"
+    skip_unless_explicit = pytest.mark.skipif(not explicit_on,
+                                              reason=f"IntegrationTestDataDeriveUtil tests skipped locally; you can "
+                                                     f"activate by setting 'DERIVE_UTIL_IT_ON' to 'true' in your "
+                                                     f"project '{TestBase._TEST_ENV_FILE_BASENAME}' file.")
+except ImportError as e:
+    raise RuntimeError(f"Failed to execute tests due to error with check of whether special skip logic applies") from e
+
+
+@skip_unless_explicit
```
aaraney, line 53: If `pytest` is installed by you run this with `unittest` what happens? I would think that this is a no-op?

aaraney (changes requested):
Looks solid, just a few comments to work through! Thanks, @robertbartel!

python/lib/core/dmod/core/dataset.py
```diff
@@ -16,6 +16,46 @@
...
+# TODO: (later) consider putting this in a different module, and also adding utils for doing the archiving
+class DataArchiving(PydanticEnum):
+    """
+    Supported archiving techniques that may be applied to the entire contents of a file-based dataset.
+
+    Types of whole-dataset archiving techniques that are supported for DMOD datasets.  Note that these represent ways to
+    archive **all** the data of a dataset, when the dataset itself requires archiving.  Datasets may also contain data
+    archive files as individual data items, and such archive files are not necessarily restricted to these types.
+    """
+    TAR = (1, ".tar")
```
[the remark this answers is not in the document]
robertbartel: Partially at least because we'd have to change the file extensions for the zip-related values; right now they are all `.zip` for simplicity.  Whether it makes sense to do that here is debatable, but my initial thought was that this was consistent with real-world usage.

python/lib/core/dmod/core/dataset.py
```diff
@@ -125,6 +165,20 @@ class Dataset(Serializable):
...
     created_on: Optional[datetime] = Field(description="When this dataset was created, or ``None`` if that is not known.")
     last_updated: Optional[datetime]
+    data_archiving: Optional[DataArchiving] = Field(default=None,
+                                                    description="Indication of what, if any, whole-dataset archiving "
+                                                                "(as opposed to simply containing an archive as one of "
+                                                                "the data files/items) is applied by dataset backend "
+                                                                "to store this dataset's data.")
+
+    @validator("data_archiving")
+    def validate_data_archiving(cls, v, values):
```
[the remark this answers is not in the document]
robertbartel: I don't think we need to make `data_archiving` private, but it would be nicer if we could (easily) encapsulate the validation with a setter instead of it happening during init.

I have switched the validator itself to a root validator to make sure we avoid issues with defaults, etc..

python/lib/core/dmod/core/dataset.py
```diff
@@ -267,6 +340,22 @@ def _set_expires(self, new_expires: datetime):
         self.expires = new_expires
         self.last_updated = datetime.now()
 
+    @property
+    def archive_name(self) -> Optional[str]:
```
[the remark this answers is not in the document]
robertbartel: In the strictest sense, probably no.  But it seems intuitive to model it as a derivative property of a _Dataset_ instance, and trivial to implement.  Is there a reason why we shouldn't have it?

I did specifically separate the static method so that things could replicate a value without having a _Dataset_ object directly, assuming the hypothetical params could be determined, though I do see even that as being more appropriate if tied to the _Dataset_ class.

python/lib/modeldata/dmod/modeldata/data/object_store_manager.py
```diff
@@ -245,6 +247,14 @@ def add_data(self, dataset_name: str, dest: str, domain: DataDomain, data: Optio
         ::method:`_push_file`
         ::method:`_push_files`
         """
+        # Prevent adding to read-only dataset except when first setting it up
```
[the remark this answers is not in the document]
robertbartel: A question for future proofing:  we don't account for it yet, but eventually, we will likely need to lock datasets from changes when they are in use.  Should this invariant be encapsulated in an isolated (perhaps layered) manner, such that we have a function that is just the read-only-not-new test, or lumped into something that will eventually grow more broad?

python/services/dataservice/dmod/dataservice/initial_data_adder_impl.py
```diff
@@ -52,6 +55,8 @@ def add_initial_data(self):
         DmodRuntimeError
             Raised when initial data could not be assembled and/or added successfully to the dataset.
         """
+        original_domain = self._dataset_manager.datasets[self._dataset_name].data_domain
```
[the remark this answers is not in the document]
robertbartel: I am on board with talking about changes to this (though separately).   I'm pretty sure we'd need both `get_dataset` and `has_dataset` functions (at least), which in some ways seems like it might be _more_ details that need to be known, but it would also give us better encapsulation and information hiding.

Aside from that, I already have some other things on my TODO list for changing how initial data adders and datasets interact, as I agree that the coupling isn't very well set up.

python/lib/core/dmod/core/dataset.py
```diff
@@ -16,6 +16,46 @@
 from .common.reader import Reader
 
 
+# TODO: (later) consider putting this in a different module, and also adding utils for doing the archiving
+class DataArchiving(PydanticEnum):
```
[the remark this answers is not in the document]
robertbartel: As mentioned elsewhere, I think we have trouble with reusing the file extensions for the ZIP variants if we do that.  We can discuss the merits of each of these further, though.

python/lib/core/dmod/core/dataset.py
```diff
@@ -16,6 +16,46 @@
...
+    """
+    Supported archiving techniques that may be applied to the entire contents of a file-based dataset.
+
+    Types of whole-dataset archiving techniques that are supported for DMOD datasets.  Note that these represent ways to
+    archive **all** the data of a dataset, when the dataset itself requires archiving.  Datasets may also contain data
+    archive files as individual data items, and such archive files are not necessarily restricted to these types.
+    """
+    TAR = (1, ".tar")
+    """ Tar archiving with no compression. """
+    TAR_GZIP = (2, ".tar.gz")
```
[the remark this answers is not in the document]
robertbartel: Sure, I went ahead and added it just now.

python/lib/modeldata/dmod/modeldata/data/object_store_manager.py
robertbartel: > This feels like we are injecting a use case specific feature into a general api ...

I disagree with the headline here but agree with parts of your argument in isolation.  

It is a use case specific feature, but it's not part of the API.  It's a specific implementation of the API that provides this behavior, but behind the scenes and particular to a subset of use cases.  I don't see a better way to introduce storage-backend-specific optimizations other than doing it within the thing interacting with the backend.  A specific application is not going to have any idea (nor should it) that there is a performance penalty for certain write scenarios (e.g., lots of files to the object store).

And, just to be clear, this is intended as a backend-specific optimization.  It's only being introduced _here_ because MinIO can take advantage of ZIP files in a particular way.  I don't have a problem with other things _also_ archiving data in a dataset in the future, if they have other reasons to do so.  I expect eventually we'll want to; hence, the attribute within _Dataset_ instead of just tracking inside the manager.   But those are apples and oranges.

But I agree that this doesn't consider all scenarios sufficiently:  a single netcdf forcing file, for example, probably doesn't need to be archive.  Probably, 50 forcing CSV files don't really either.  I even put in a TODO comment to this effect, but didn't want to just throw out a magic number of files for the too-many threshold.  I'm open to discussing either what that number should be or how to better determine when the object store manager really can/should do this.

> One thought is to create an algebraic datatype like a DataFormat.ARCHIVE[ArchiveFormat, DataFormat] that is, itself, a DataFormat but also wraps an ArchiveFormat and a DataFormat

That seems much more complicated and far reaching that what's in place now.  And I still don't think we could effectively apply it without making the things creating the data start to worry too much about the details of the data storage backend\*.  Not to mention it would convolute logic for reconciling data requirements.  

\* Caveat that this is already the case for workers writing output, kind of, though they don't interact with the DMOD data orchestration in the same way, so their data writing is written more in isolation (at least for now).

python/services/dataservice/dmod/dataservice/initial_data_adder_impl.py
robertbartel: Perhaps, but this is probably too far out of scope here.  Admittedly, I did modify calls to `self._dataset_manager.add_data` in the _InitialDataAdder_ implementations.  One could argue (questionably) that this bled into scope due to it being done as part of the other modifications to _BmiAutoGenerationAdder_, but, more importantly, those changes were to fix something that was broken.

python/services/dataservice/dmod/dataservice/initial_data_adder_impl.py
robertbartel: Good catch.  I've pushed a change for it.

python/services/dataservice/dmod/test/it_data_derive_util.py
robertbartel: I've made a modification to how this is handled that should handle that situation more appropriate.

python/services/dataservice/dmod/dataservice/initial_data_adder_impl.py
aaraney: I think you could argue this is out of scope, but IMO the change is minor and the potential performance consequences are high. My concern is if a large file is passed as `path`. This will blow up the resident memory usage of the process b.c. the file read is greedy (likely a demand paged `mmap` that is `munmap`ed, but still not great). If we just pass something that has a `read()` method, `add_data` can perform buffered reads that have a lesser potential to degrade performance.

aaraney (commented):
Really just one major comment that I think is worth addressing and then we should be good to go! Thanks, @robertbartel!

robertbartel:
@aaraney, I've fixed conflicts and I think addressed your last concern by passing a reader instead of the raw bytes.  Let me know if there is anything else.

aaraney (approved):

aaraney:
Looks like after this was merged the [integration tests](https://github.com/NOAA-OWP/DMOD/actions/runs/10253857745/job/28410913531) are now failing on master. I had a hunch this might be a caching issue. So, I cleared the runner caches but that did not fix the issue. Trying to reproduce locally now.

aaraney:
The service packages [were not being installed](https://github.com/NOAA-OWP/DMOD/blob/452bf58b018955ff8cf6a59abb326d04232ce35f/.github/workflows/integration_tests.yml#L77) in the IT tests. Same fix as in #575 but for IT tests.

aaraney:
Opened https://github.com/NOAA-OWP/DMOD/pull/696 to fix this.

Issue #673, opened 2024-07-12
DMOD image build failures due to t-route


robertbartel:
Currently it is not possible to build the DMOD ngen worker image.  One of two different failures will occur, depending on whether the environment configuration is set to build the image with or without t-route.

*UPDATE*: the second failure has now been moved to its own issue: #676.

With t-route included by the image (which is the default), the following error is seen in the `rocky_build_troute` image build stage:

```bash
...
70.99 removing build/bdist.linux-x86_64/wheel
71.01 Successfully built troute_config-0.0.3.tar.gz and troute.config-0.0.3-py3-none-any.whl
71.04 python3: can't open file '/ngen/t-route/src/troute-nwm/setup.py': [Errno 2] No such file or directory
------
failed to solve: process "/bin/sh -c if [ \"${NGEN_WITH_ROUTING}\" == \"ON\" ]; then       export FC=gfortran ;       export F90=gfortran ;       mkdir ${WORKDIR}/t-route/wheels       && cd ${WORKDIR}/t-route       && ./compiler.sh       && cd ${WORKDIR}/t-route/src/troute-network       && python3 setup.py --use-cython bdist_wheel       && cp dist/*.whl ${WORKDIR}/t-route/wheels/       && cd ${WORKDIR}/t-route/src/troute-routing       && python3 setup.py --use-cython bdist_wheel       && cp dist/*.whl ${WORKDIR}/t-route/wheels/       && cd ${WORKDIR}/t-route/src/troute-config       && python3 -m build .       && cp dist/*.whl ${WORKDIR}/t-route/wheels/       && cd ${WORKDIR}/t-route/src/troute-nwm       && python3 setup.py bdist_wheel       && cp dist/*.whl ${WORKDIR}/t-route/wheels/ ;     fi" did not complete successfully: exit code: 2
Error: requested 'build' action failed (17); exiting without proceeding further
```

This is likely due to recent changes in t-route, though confirmation is still needed.

robertbartel:
@aaraney, I don't suppose you can tell quickly from this exactly what needs to be adjusted in the image's t-route build stage, can you?

aaraney:
[...]

PR #675, opened 2024-07-12
Update worker images to optimize IO performance using local data

robertbartel:
*Note*: do not review until #671 is complete.
*Note*: blocked by #673, as testing cannot yet be completed.

Optimizing job execution by sometimes locally copying DMOD dataset data to local, per-node Docker volumes when jobs are starting.

## Changes

- Update _DataRequirements_ to track if the data is required to be local with new `needs_data_local` attribute
- Updating usages of _DataRequirements_ to ensure `needs_data_local` is set whenever `fulfilled_by` is set (i.e., de facto couple this as a part of fulfillment details)
- Add new directory structures within worker image for mounted datasets and for local data, and have image symlink into previous directory (`/dmod/datasets/**`) at startup
- Update scheduler _Launcher_ to create local volumes in job services for those requirements needing local data, in addition to the volumes mounting the dataset-backed data
- Update worker images to include the MinIO CLI client for optimal copying when making data local (10x speedup)
- Updating worker entrypoints and scripts to support detecting when data needs to be local by the prescence of a mounted local volume corresponding to the dataset volume, and then making the data local (one worker per physical node) on that node

## Testing

Pending; blocked by #673.


docker/main/ngen/Dockerfile
```diff
@@ -906,8 +908,16 @@ COPY --chown=${USER} --from=build_bmi_sac_sma /dmod/ /dmod/
...
 USER root
-# Update path and make sure dataset directory is there
-RUN echo "export PATH=${PATH}" >> /etc/profile \
+# TODO: (later) consider something like this in the future (at least optionally) when downloading the mc client below:
+# ARG MINIO_CLIENT_RELEASE="RELEASE.2024-07-08T20-59-24Z"
+# ...
+# https://dl.min.io/client/mc/release/linux-amd64/archive/mc.${MINIO_CLIENT_RELEASE}
+
+# Setup minio client; also update path and make sure dataset directory is there
+RUN curl -L -o /dmod/bin/mc https://dl.min.io/client/mc/release/linux-amd64/mc \
```
aaraney, line 917:
```suggestion
RUN ARCHITECTURE=`echo $(uname -s)-$(uname -m) | tr '[:upper:]' '[:lower:]'`; \
curl -L -o /dmod/bin/mc https://dl.min.io/client/mc/release/${ARCHITECTURE}/mc \
```

docker/main/ngen/Dockerfile
```diff
@@ -906,8 +908,16 @@ COPY --chown=${USER} --from=build_bmi_sac_sma /dmod/ /dmod/
...
+    && chmod +x /dmod/bin/mc \
+    && mkdir /dmod/.mc \
+    && echo "export PATH=${PATH}" >> /etc/profile \
```
aaraney, line 920: Any reason not to do:

```suggestion
    && echo "export PATH=${PATH}"':$PATH' >> /etc/profile \
```

docker/main/ngen/Dockerfile
robertbartel: Hmm ... good catch, but the suggestion won't work as needed with Linux on an Intel machine (it produces `linux-x86_64` instead of `linux-amd64`).  I will make an adjustment, but I'll need to think more about exactly how.

docker/main/ngen/Dockerfile
robertbartel: Is it really a good idea to allow the environment state after loading the global profile to be dependent on the prior environment state like this?  And aside from whether it's safe, is it actually useful in practice?

docker/main/ngen/Dockerfile
robertbartel: That wasn't the only case that would have been a problem, but I've made a change now that I think should catch the platform types we can reasonably expect that would need to be transformed to something else for purposes of that URL, and done the adjustment accordingly.

docker/main/ngen/Dockerfile
aaraney: I must have missed the above line in my initial review:

```shell
ENV PATH="$VIRTUAL_ENV/bin:$PATH"
```

I think this is a moot point.

Issue #661, opened 2024-06-11
Update existing or develop new tools for more granular control of tags when building worker images 


robertbartel:
Support better control and management of worker images via use of more tagging. 

- Generate and utilize tag values by default that reflect the current state of the image configuration
  - Generate a tag to use for "normal" scenarios that is based on the current version/commit of DMOD itself
  - Dynamically alter the tag for cases when Dockerfile or build context of local repo has been modified, deterministically deriving something for use in the tag from the altered state
  - Dynamically alter the tag for cases when the local DMOD environment config file specifies a different commit, branch, or repo than "standard", again deterministically deriving something reflected of the differences
  - Reference the baseline "normal" tag details in these dynamically altered versions
  - Have something in the dynamically altered versions to indicate the type of situation leading to the alteration (e.g., maybe include `rc` somewhere in the tag when there are local changes in the repo) 
- Allow control when building to completely override the generated param in favor of something user-provided

robertbartel:
@christophertubbs, I'm moving the relevant thread of the conversation from #651 over here.

> ... I still think it'd be wise to append usernames as tags to built images to maintain image uniqueness in shared environments and allow common names as values from optional parameters - i.e. go with the untagged/"official" version when told to, but default to 127.0.0.1:5000/dmod-data-service:$USER.

I think it would be better to create tags deterministically based on the state of the build context (i.e., the repo and the .env config), simply based on DMOD commit id (and/or eventually git tags) for "standard" situations, and adjusting tag values to reflect when/how the context changes.  Along with, of course, providing the option to manually override all this.   It gives more granular control over and tracking of previous versions, especially when developing the images.

That said, I'm pretty sure there isn't anything stopping us from doing _both_.   I don't see any reason that we couldn't also apply a tag equal to the username every time a user builds a new image (though we might want to provide an .env config option to turn this off).  We would have to account for the fact that the username approach requires re-use of existing tags, whereas the state-based approach _should_ avoid that, but I'm sure that can be worked out.

PR #678, opened 2024-07-17
Add controls for numpy version

robertbartel:
~Add support for environment-config-controlled constraints on the version of numpy installed by the image build process for ngen-related Docker images.~ (**Note**: scrapping original idea in favor of a different approach.)

Add support for pip-constraints-file based control over the version of numpy installed by the image build process for ngen-related Docker images.

Slightly related to noaa-owp/ngen#837.

Note that this blocks #675, as that revision of the image can't be built for testing due to the problem being fixed here.

## Additions

- ~New environment config variable for numpy constraint, with default value and comments in _example.env_~
- New _constraints.txt_ file defining pip constraints for ngen (and related) image builds.

## Removals

- Removes a deprecated and unused Dockerfile `ARG` `MIN_NUMPY` for clarity.

## Changes

- ~New Dockerfile `ARG` for numpy constraint, applied appropriately where numpy gets installed by image build~
- ~Setting of aforementioned `ARG` in Compose build yaml configs, for all ngen-based images/services (e.g., model exec, calibration, etc.)~
- `COPY` and `ENV` statements in ngen image Dockerfile to bring in new pip constraint and then set `PIP_CONSTRAINT` environment variable so that pip uses it.

## Testing

1.  ngen worker image builds successfully, whereas previously it would fail due to a failed ngen unit test testing part of the Python BMI integration


docker/main/docker-build.yml
```diff
@@ -49,6 +49,7 @@ services:
       context: ./ngen
       target: rocky-ngen-deps
       args:
+        NUMPY_CONSTRAINT: ${NUMPY_INSTALL_CONSTRAINT:?Please set global NUMPY_INSTALL_CONSTRAINT in environment config}
```
christophertubbs, line 52: This is an abuse of the `args` parameter. Do not use this to inject python versions. Stick to `setup.py` and `requirements.txt`.

docker/main/docker-build.yml
```diff
@@ -59,6 +60,7 @@ services:
       context: ./ngen
       target: rocky_ngen_build_testing
       args:
+        NUMPY_CONSTRAINT: ${NUMPY_INSTALL_CONSTRAINT:?Please set global NUMPY_INSTALL_CONSTRAINT in environment config}
```
christophertubbs, line 63: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

docker/main/docker-build.yml
```diff
@@ -77,6 +79,7 @@ services:
       context: ./ngen
       target: ngen_worker
       args:
+        NUMPY_CONSTRAINT: ${NUMPY_INSTALL_CONSTRAINT:?Please set global NUMPY_INSTALL_CONSTRAINT in environment config}
```
christophertubbs, line 82: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

docker/main/docker-build.yml
```diff
@@ -98,6 +101,7 @@ services:
       context: ./ngen
       target: partitioner_image
       args:
+        NUMPY_CONSTRAINT: ${NUMPY_INSTALL_CONSTRAINT:?Please set global NUMPY_INSTALL_CONSTRAINT in environment config}
```
christophertubbs, line 104: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

docker/main/docker-build.yml
```diff
@@ -114,6 +118,7 @@ services:
       context: ./ngen
       target: ngen_cal_worker
       args:
+        NUMPY_CONSTRAINT: ${NUMPY_INSTALL_CONSTRAINT:?Please set global NUMPY_INSTALL_CONSTRAINT in environment config}
```
christophertubbs, line 121: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

docker/main/ngen/Dockerfile
```diff
@@ -65,9 +65,11 @@ ARG BOOST_VERSION=1.82.0
 #mpich 3.2 doesn't work well gfortran 11 it seems, an alignment error crops up, but 3.3.2 seems to work...
 ARG MPICH_VERSION="3.3.2"
 ARG MIN_PYTHON="3.8.0"
-ARG MIN_NUMPY="1.18.0"
 ARG BLOSC2_VERSION
 
+# Necessry to control specific numpy version after release of 2.0 and prior to fully support by all NextGen ecosystem
+ARG NUMPY_CONSTRAINT
+
```
christophertubbs, line 72: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

docker/main/ngen/Dockerfile
```diff
@@ -290,14 +292,15 @@ RUN if [[ "${NETCDF_CXX_VERSION}" == "latest" ]]; then \
 FROM rocky-base as rocky-ngen-packaged-deps
 
 ARG ROCKY_NGEN_DEPS_REQUIRED
+ARG NUMPY_CONSTRAINT
```
christophertubbs, line 295: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

docker/main/ngen/Dockerfile
```diff
@@ -290,14 +292,15 @@ RUN if [[ "${NETCDF_CXX_VERSION}" == "latest" ]]; then \
...
 
 # TODO: later, go back and change all pip3/python3 to just pip/python (but leave for now to limit scope)
 # Note that this includes numpy, which is needed for Python BMI support, regardless of BMI module 
 USER root
 RUN dnf update -y && dnf install -y ${ROCKY_NGEN_DEPS_REQUIRED} && dnf clean -y all \
     && ln -s $(which python3) $(which python3 | sed 's/python3/python/') \
     && pip install --no-cache-dir "pip>=23.0,<23.1" wheel packaging \
-    && if [ "${NGEN_WITH_PYTHON}" == "ON" ]; then pip install --no-cache-dir numpy; fi
+    && if [ "${NGEN_WITH_PYTHON}" == "ON" ]; then pip install --no-cache-dir numpy${NUMPY_CONSTRAINT}; fi
```
christophertubbs, line 303: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

docker/main/ngen/Dockerfile
```diff
@@ -652,6 +654,8 @@ ARG BUILD_TOPMODEL
 ARG BUILD_PET
 ARG BUILD_SLOTH
 
+ARG NUMPY_CONSTRAINT
```
christophertubbs, line 657: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

docker/main/ngen/Dockerfile
```diff
@@ -666,8 +670,9 @@ RUN if [ "${NGEN_WITH_PYTHON}" == "ON" ]; then \
 USER ${USER}
 
 ENV VIRTUAL_ENV=/dmod/venv
-RUN python3 -m venv $VIRTUAL_ENV && pip3 install numpy
+RUN python3 -m venv $VIRTUAL_ENV
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+RUN if [ "${NGEN_WITH_PYTHON}" == "ON" ]; then pip install numpy${NUMPY_CONSTRAINT} ; fi
```
christophertubbs, line 675: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

docker/main/ngen/Dockerfile
```diff
@@ -783,7 +788,25 @@ FROM rocky-ngen-deps as rocky_ngen_build_testing
 #FROM rocky_build_ngen as rocky_ngen_build_testing
 #FROM ${DOCKER_INTERNAL_REGISTRY}/ngen-deps:latest as rocky_ngen_build_testing
 
+ARG NUMPY_CONSTRAINT
```
christophertubbs, line 791: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

docker/main/ngen/Dockerfile
```diff
@@ -783,7 +788,25 @@ FROM rocky-ngen-deps as rocky_ngen_build_testing
...
+RUN if [ "${NGEN_WITH_PYTHON}" == "ON" ]; then \
+        chgrp -R ${USER} /usr/local/lib*/python3.* ; \
+        chmod -R g+sw /usr/local/lib*/python3.* ; \
+    fi
+USER ${USER}
+
+ENV VIRTUAL_ENV=/dmod/venv
+RUN if [ "${NGEN_WITH_PYTHON}" == "ON" ]; then python3 -m venv ${VIRTUAL_ENV} ; fi
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+RUN if [ "${NGEN_WITH_PYTHON}" == "ON" ]; then pip install numpy${NUMPY_CONSTRAINT} ; fi
```
christophertubbs, line 807: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

example.env
```diff
@@ -192,6 +192,21 @@ PYCHARM_REMOTE_DEBUG_SERVER_PORT_GUI=55875
...
+
+## Version of Numpy to select when installing (e.g., within Docker images)
+## Note that this probably needs to stay in sync with what is in a few other places
+##  that don't read directly from this file:
+##      - requirement.txt
+##      - python/gui/dependencies.txt
+# Note: mind the consecutive equal signs; the value being set here is '==1.26.4'
+NUMPY_INSTALL_CONSTRAINT===1.26.4
+
+
```
christophertubbs, line 209: This is an abuse of the args parameter. Do not use this to inject python versions. Stick to setup.py and requirements.txt.

The use of the `.env` pattern here needs to be reduced, not added to.

christophertubbs (dismissed):

docker/main/docker-build.yml
robertbartel: We don't exactly have a requirements.txt for _inside_ the worker images (they kind of use the one that's part of the ngen Python BMI tests, but that's not really what that is for).  I suppose we could go that way, but this then would have to become a subtle overhaul of what we are doing inside the images to install numpy, rather than an extension to what's already in place.  I.e., the scope, size, and impact would change significantly.  At that point, I think we'd have to at least reconsider whether to just wait for the numpy 2.0.0 compatibility issue to be fixed on the ngen side.

Aside from that, this also would have the side effect of committing to a particular numpy version in the code, rather than allowing users to configure that per deployment.  I'd be interested in hearing an argument as to whether that in and of itself is a preferable approach, if you're suggesting that.

python/gui/dependencies.txt
```diff
@@ -17,4 +17,4 @@ channels
 channels-redis
 djangorestframework
 psycopg2-binary  # TODO: get source package in future. Note that psycopg2 cannot be used on Mac; psycopg2-binary must be used
-numpy
+numpy==1.26.4
```
aaraney, line 20: No need to be so specific with the version. We just want to ensure that it is below 2.
```suggestion
numpy~=1.0
```

example.env
robertbartel: > The use of the .env pattern here needs to be reduced, not added to.

I don't think I agree with this exactly as intended.  Our structures for configuring a DMOD environment need to be improved; using .env files like this is not the most effective thing we could be doing.  But that shouldn't equate to reducing what users are able to configure, or restricting ourselves from adding more configurable options when they are needed.

But, setting aside how we implement things, should we simply not allow users to control what version of numpy is installed in worker images?  If not, why?

requirements.txt
```diff
@@ -20,7 +20,7 @@ python-dotenv
 djangorestframework
 git+https://github.com/noaa-owp/hypy@master#egg=hypy&subdirectory=python
 hydrotools.nwis-client
-numpy
+numpy==1.26.4
```
aaraney, line 23: I think adding this alone should resolve our issue. If a dependency requires `numpy>1`, things will start failing, but I don't think the arguments added above will address that either. I think it makes the most sense to add this check into `ngen` for the time being until support is added. That way, `ngen` just fails if it detects that `numpy>1` is being used. 

```suggestion
numpy~=1.0
```

example.env
christophertubbs: If users desire to change versions of library dependencies they are coloring that far outside the lines they need to do so via channels like their requirements.txt, setup.py files, or their own versions of their build scripts and files.  This degree of configuration is too much and provides too many avenues for failure.

aaraney (changes requested):
I think was should instead add a `requirements.txt` for the `ngen` image and put things like this there. I think that will easiest to maintain and understand in the long run. I think it also results in a simpler solution. Thoughts?

example.env
robertbartel: > If users desire to change versions of library dependencies they are coloring that far outside the lines they need to do so via channels like their requirements.txt, setup.py files, or their own versions of their build scripts and files.

I worry there will be a particular problem here when we start to bring in BMI modules that aren't directly integrated into ngen or the DMOD image (which is coming soon).  For a users to essentially dynamically inject a BMI module for use within job workers, it's very likely that at least some of them will need the ability to control the version of numpy installed (i.e., when ngen is compiled).  

We could, alternatively, disallow (formally or de facto) any injected modules if they don't support a numpy version hard-coded into the worker.  In principle, I think that hurts community adoption; in practice, maybe the scenario doesn't happen frequently enough to outweigh other concerns.

python/gui/dependencies.txt
robertbartel: I'm strongly considering taking this out entirely.  I realized there really is no evidence at this point that there is a problem with version 2.0.0 for anything running inside the GUI stack.  The point was to match the default I was setting up, but even assuming we still went that route, that really isn't necessary.

requirements.txt
robertbartel: I'm strongly leaning toward taking this out also.  Again, the real need was for ngen compilation inside the workers.  This goes beyond that.  Further, this file actually applies to the place where the user actually has (and likely is capable of handling) the most control over what numpy is installed.  I don't think we want to take that control away.

aaraney: Yeah, sorry I wasn't paying the closest attention. I would advocate for creating a ngen worker `requirements.txt` (if needed) that we copy into the worker `dockerfile` and install if python is enabled. I think that is the easiest solution. Thoughts?

example.env
christophertubbs: As it stands, the requirements of "env file + environment variables + multiple Dockerfiles + Swarm" will hurt adoption on its own. 

What are the plans for dynamically building images with specific numpy versions if the the lever for the setting is global for the entire runtime environment? Is it going to require direct calls to `docker build` with a long list of `arg` parameters? If so, does that also indicate that the process of building images has _also_ become too cumbersome?

When these requirements appear, a solution for that needs to be implemented using different strategies, such as adding `COPY` hook into the Dockerfile to allow for more advanced behavior for more advanced BMI modules. That, in of itself, is easier than juggling `ARGS` within the image definitions.

docker/main/docker-build.yml
robertbartel: > I'd say it might be up for debate for us to really set a version - the installation of libraries like pandas will do that for us.

Eh, there isn't a way around us doing this, at least implicitly, because ngen needs to be compiled with some version of numpy before anything like that is installed.

> Relying on args in this context is an indication that we are doing something wrong. If this is the only solution, we need to take a step back and fix the mistakes that require this.

It isn't the only way to fix this problem, but it's the only way I see to fix it by providing dynamic user control over the numpy version, and there at least an argument for needing that when we dynamically inject user-provided BMI modules into job workers.

> Additionally, if ngen is broken, it is not DMOD's responsibility to fix it.

There's a debate there too.  ngen still works fine as long as you don't install numpy 2.0.  It's the DMOD images that are doing that.

christophertubbs:
If the issue is with a single requirements file... there's nothing keeping us from having a single requirements file. We can also produce new build 'hooks' where extra files may be COPYed in if behavior needs to be altered in different stages.

What we have, right now, is already too complicated. We need to start moving in another direction to find an easier to follow build pipeline. If _not making something configurable_ makes the build process prohibitively resource intensive, something is wrong and we need to take a step back and evaluate our process(es).

For now, hard code the value since it would be global anyways. If a user needs to color outside the lines and use a different value, they may change it themselves.

Ideally, DMOD needs to get out of the game of setting up compilation rules for another product. We don't set compilation flags or environment variables just to prepare pandas for use. DMOD is a consumer of NextGen and compatible formulations. Instructions and constraints for use need to be provided _from_ them, not developed by us. If the DMOD image creation process is responsible for setting this up we're putting ourselves in a position of eternal catch up.

example.env
robertbartel: > As it stands, the requirements of "env file + environment variables + multiple Dockerfiles + Swarm" will hurt adoption on its own.

I'd be shocked if one more configurable .env setting hurt adoption any more than using that approach for deployment configuration to begin with.  And I'm not suggesting we don't overhaul how deployments are configured, but that's way out of scope for this.

> What are the plans for dynamically building images with specific numpy versions if the the lever for the setting is global for the entire runtime environment?

I'm not really sure what you are asking here.  If you're talking about ways to customize different tags of the worker image with potentially different versions of numpy, I wasn't going to get into that here.  This was just to fix something that was broken (with the solution being a new knob to turn, rather than something hard-coded, since I expected we might need that knob soon anyway for other reasons).  But I'm sure once we start to get into #661 we'd need to think about things like this.

robertbartel:
> What we have, right now, is already too complicated. We need to start moving in another direction to find an easier to follow build pipeline.

You aren't wrong, but that's out of scope for what is essentially a bug fix.

> If not making something configurable makes the build process prohibitively resource intensive, something is wrong and we need to take a step back and evaluate our process(es).

Just to be clear, the lack of configurability doesn't make the builds less efficient.  But artificially tethering together the installation of all the Python packages used throughout the worker images would.

> For now, hard code the value since it would be global anyways. If a user needs to color outside the lines and use a different value, they may change it themselves.

Sure, I can make that change for here, and we can separate off the idea of adding (or not) configurability to numpy versioning for workers.

> Ideally, DMOD needs to get out of the game of setting up compilation rules for another product. We don't set compilation flags or environment variables just to prepare pandas for use. DMOD is a consumer of NextGen and compatible formulations.

Providing compatible, consistent, reproducible compute environments (whether these be Docker images/containers or some other mechanism) for model execution is a fundamental feature of DMOD, and that means it has to be concerned with compilation flags and environment details and anything else users would have to deal with to set up an environment if they were just trying to use ngen directly.  There is certainly room to discuss which of these DMOD opens up via control/configure mechanisms versus what DMOD just handles automagically behind the scenes.  And if/when ngen can add its own automation for installing its dependencies like numpy or begins distributing installation packages, great; DMOD can leverage those instead.  Until it does though, for DMOD to be able to run ngen, there’s just no way around dealing with this kind of stuff.

aaraney:
I think https://github.com/NOAA-OWP/ngen/pull/856 will solve our issues an obsolete this PR. We will still at some point need to revisit how we handle external bmi module dependency versioning, but that is a whole can of worms that I think we should kick down the road!

robertbartel:
@aaraney, no, I think #https://github.com/NOAA-OWP/ngen/pull/856 will just cause DMOD image builds to fail for a different reason (the CMake error instead of the unit test failure).  Unless ngen actually handles installing numpy itself, we are going to have to address what version of numpy gets installed in the image.

As discussed above, I'll make changes so that we hard code this for now to the latest 1.x.

aaraney:
Extra context, there is already a [check](https://github.com/NOAA-OWP/ngen/blob/0917f9b021e081f01d42966ad92ee6370f3eca54/include/utilities/python/InterpreterUtil.hpp#L114-L119) in `ngen` that if the version of `numpy` that was used at build time does not match the one that is used at runtime, an exception is thrown. So, if `ngen` is built with `numpy<=2` and then `numpy` in the environment gets updated, we are still covered.

aaraney:
Alright, so after a side channel conversation with @robertbartel, at least IMO, I think we can achieve the goals of this PR using a `pip` [`constraints.txt`](https://pip.pypa.io/en/stable/user_guide/#constraints-files) file (very similar to `requirements.txt`). More specifically, the path to the `constraints.txt` file would be set to the [`PIP_CONSTRAINT`](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-c) environment variable in any build stages that build / install python dependencies. In doing so, any packages that are installed via `pip` will be constrained to the version specifiers in `constraints.txt`.

In doing so, this lets us control what version or version ranges of certain packages are installed in the image. Likewise, incompatibilities will results in a build failure.

docker/main/ngen/Dockerfile
```diff
@@ -290,14 +292,15 @@ RUN if [[ "${NETCDF_CXX_VERSION}" == "latest" ]]; then \
 FROM rocky-base as rocky-ngen-packaged-deps
 
 ARG ROCKY_NGEN_DEPS_REQUIRED
+ARG NUMPY_CONSTRAINT
 
 # TODO: later, go back and change all pip3/python3 to just pip/python (but leave for now to limit scope)
 # Note that this includes numpy, which is needed for Python BMI support, regardless of BMI module 
 USER root
```
aaraney, line 299: I figure we could move the _required_ python packages to a `requirements.txt` file (so, `pip`, `wheel`, `packaging`, and `numpy`) and then create a `constraints.txt` that we use to enforce the allowed versions of the required and other packages in `constraints.txt`. Thoughts?

```suggestion
USER root
COPY requirements.txt ${WORKDIR}/requirements.txt
COPY constraints.txt ${WORKDIR}/constraints.txt
```

docker/main/ngen/Dockerfile
aaraney, line 303: This would become:

```suggestion
    && if [ "${NGEN_WITH_PYTHON}" == "ON" ]; then pip install --no-cache-dir --requirement requirements.txt --constraint constraints.txt ; fi
```

docker/main/ngen/Dockerfile
christophertubbs: Good find! I think it's a good solution.

It's been a while since I've had to worry about this, but I _believe_ that editing the `constraints.txt` and creating a new hash for the file would cause a new `docker build` to start at the line for the `constraints.txt` file , so I think we can be sure that a change there will just about always be reflected after a build.

robertbartel: I still don't like the idea of coupling all the Python packages together in a _requirements.txt_, as I don't think that fits with how the Dockerfile is set up.  It's totally fair to discuss whether _that_ needs to change - e.g., should we _really_ support building the image both with or without Python support (we don't really need numpy if in the without option)) - but it needs to be discussed separately.

But the `PIP_CONSTRAINT` and _constraints.txt_ settings sound fine.  Those alone should be sufficient to resolve the immediate issue.

robertbartel: @christophertubbs, you raise a good point I hadn't consider:  depending on when the _constraints.txt_ file is setup in the image that will once again have implications on build caching and efficiency.  That isn't an absolute blocker, but I want to think on it a little more to see if there's anyway to avoid that tradeoff.

aaraney: > I still don't like the idea of coupling all the Python packages together in a _requirements.txt_, as I don't think that fits with how the Dockerfile is set up.

Yeah, I agree. Im just talking about the required packages to build `ngen` with python support. So, `numpy` for now but this could expand to others in the future.

docker/main/ngen/constraints.txt
```diff
@@ -0,0 +1,5 @@
+### Constrain pip
+pip>=23.0,<23.1
```
aaraney, line 2: I don't think this is necessary after https://github.com/NOAA-OWP/t-route/pull/804 was merged.

aaraney: Doesn't have to be addressed here, just putting it on your radar @robertbartel!

aaraney (approved):
This looks good to me! Thanks, @robertbartel!

robertbartel:
@christophertubbs, any other concerns before this moves forward?  I might either need explicit approval from you or to dismiss your previous review, in order to comply with branch protections (the latter of which I don't want to do until you've had a chance to raise any remaining discussion items).

PR #681, opened 2024-07-18
Use `ScalarType` instead of `sctypes` when attempting to tie data types to their names

### LINKED ISSUE #679: NumPy 2.0 Breaks `type_name_to_dtype` transformations in evaluations
Users are able to dictate what type of data will be in a column if they seek to cast, so stuff like announcing "This column is filled with dates" or "These are floats" or "These are strings" (important for numeric identifiers).

99% of the use cases of this are covered by simple checks in `dmod.evaluations.util.type_name_to_dtype`, but detailing anything other than "date", "datetime", "int"/"integer", "float", "bytes", "string"/"str"/"word"/"words" causes the function to look up names in `numpy.sctypes` which is being removed.

Instead of looking in `numpy.sctypes` to find what to convert to, look in `numpy.ScalarType` which is a tuple of all support dtypes.

christophertubbs:
closes #679 

`numpy.sctypes` is going away in Numpy 2.0 and `dmod.evaluations.util` relies on it. This switches up the implementation to use the tuple of dtypes instead of the old map.


aaraney (approved):
Look great! Thanks, @christophertubbs!

PR #695, opened 2024-08-02
First set of fixes to revive Django-based GUI code

robertbartel:
A set of changes to get older, Django-based GUI components for map and job configuration at least mostly loading and rendering properly, whereas before nothing or almost nothing would appear.

A few things in particular to point out:
- Fix _daphne_ dependency
- Fix lack of automated running of Django migrations (broke startup for "new" GUI deployment)
- Fix broken config for static files
- Minimally fix GUI classes that were not up to date with _dmod.communication_ and had required us to comment out some configured URLs

This is likely the first of several changes necessary to get things functional again.


python/gui/MaaS/templates/base.html
```diff
@@ -1,12 +1,12 @@
...
 <!DOCTYPE html>
 <html lang="en">
     <head>
         <meta charset="UTF-8">
         <title>{{ title|default:"Distributed Model on Demand" }}</title>
-        {% load static %}
         <link rel="shortcut icon" type="image/png" href="{% static 'common/img/favicon.png' %}"/>
         <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
-        <link rel="stylesheet" href="{% static 'common/css/base.css' %}"/>
+        <link rel="stylesheet" href="{% static 'css/master.css' %}"/>
```
christophertubbs, line 9: be careful - this is mixing and matching resources. `css/master.css` is the stylesheet for the root level `master.html` template. `common/css/base.css` is the stylesheet for `base.html`.

robertbartel: I think in this case that is what I wanted to do here, though I appreciate the cautious eye.  I felt like we should start from the same styling throughout, so I was trying to move to that in this template base.  That said, I could be talked into deferring on this if we think it's problematic to do it until it's also time to clean up some of the redundant files (which I was avoiding at the moment).

christophertubbs: You might be better off changing to base template to `master.html` on the other templates rather than altering what css `base.html` is referencing. I haven't made the switch, so it might be fine, but both css files and templates are making assumptions based on what page elements are present.

You have the right idea, though. Switching what template you're inheriting from might work better here, though.

robertbartel: Yeah, if I'm remembering correctly, I started off by switching the base template, but moving from _base.html_ caused some problem with pages' scripts.  Swapping in the CSS change ended up being easier to follow along with and appeared to have fewer immediate side effects than carefully tweaking what seemed to be a good chunk of the Javascript setup (and hoping there wasn't something else needed after that).   

Regardless, there isn't going to be a clean fix until we overhaul a good chunk of things.  My goal at moment is to be minimalist and do just enough to get things working again, albeit with some attention paid to not laying any seriously traps for us to fall into later.

christophertubbs (approved):
Good work - I look forward to seeing this evolve over time.

PR #684, opened 2024-07-29
Move from `setup.py` to `pyproject.toml`; Update build scripts

aaraney:
Move from `setup.py` to `pyproject.toml`.
Sub-package authorship is also updated.
Build scripts are also updated.

Replaces #683.
Closes #683.

Related to:
- #680
- #248
- #247 

Closes #248
Closes #247

PR #697, opened 2024-08-06
Fix some leftover references to setup.py

robertbartel:
There were a few places where references/usages/checks of a _setup.py_ file were still in place after removing those in #684.  In particular, building the _py-sources_ Docker image was broken.  This updates those spots.


aaraney (approved):

aaraney:
Thanks for the fixes, @robertbartel!

PR #700, opened 2024-08-13
Add simple Django based domain selection view

robertbartel:
Addition of a simple form-based domain selection view allowing for selecting from the available hydrofabrics and the available catchments in the selected hydrofabric before transitioning to the formulation configuration view.

Depends upon #695 so should remain a draft until that is resolved.

image

image

image

image

image


python/gui/MaaS/templates/maas/domain.html
```diff
@@ -0,0 +1,118 @@
...
+    <input type="hidden" name="feature-ids" />
+    <input type="hidden" name="fabric" />
+
+    {% for pane_input in pane_inputs %}
+        <input type="hidden" name="{{ pane_input.id }}" id="{{ pane_input.id }}" >
+    {% endfor %}
+</form>
+
+<script>
+    startup_scripts.push(loadFabricNames);
```
christophertubbs, line 113: I love seeing this get used. Good one.

python/gui/MaaS/static/common/js/domain.js
```diff
@@ -0,0 +1,206 @@
...
+        if (opt.selected) {
+            opt.selected = false;
+        }
+        selected.removeChild(opt);
+        insertOptionInOrder(opt, choices);
+    }
+}
+
+function loadFabricDomain(event) {
+    var name = $("#fabric-selector").val(),
```
christophertubbs, line 142: Use `let` instead of `var` and try to keep one declaration per statement.

```javascript
let name = ...;
let type = ...;
let catLists = ...;
```

Looks heavy, but the cognitive load is lower and it makes it easier to edit later on.

python/gui/MaaS/static/common/js/domain.js
```diff
@@ -0,0 +1,206 @@
...
+
+    $("input[name=fabric]").val(name);
+
+    // Clear any existing <option> tags from within "domainChoices" <select>
+    for (l = 0; l < catLists.length; l++) {
+        select = catLists[l];
+        for (i = select.options.length - 1; i >= 0; i--) {
+            select.remove(i);
+        }
+    }
```
christophertubbs, line 162: Try something like:

```javascript
for (let catchmentSelect of catLists) {
    for (let optionIndex = catchmentSelect.options.length - 1; optionIndex >= 0; optionIndex--) {
        catchmentSelect.remove(optionIndex);
    }
}
```

I may just be easier to do `$("#domainChoices option, #domainSelections option").remove()`, though.

python/gui/MaaS/static/common/js/domain.js
```diff
@@ -0,0 +1,206 @@
+
```
christophertubbs, line 1: Something I'm seeing that you should try is to use `let` instead of `var`, declare one variable per statement, avoid acronyms (such as `i` and `l`), use `for (let x of y)` statements instead of classical for loops, and maybe take advantage of jQuery more when it comes to iterating over collections of stuff like `option`s.

Don't sleep on `.forEach` on lists either. Super useful.

python/gui/MaaS/static/common/js/domain.js
```diff
@@ -0,0 +1,206 @@
...
+function titleCase(str) {
+    return str.replaceAll("_", " ").toLowerCase().split(' ').map(function(word) {
+        return word.replace(word[0], word[0].toUpperCase());
+    }).join(' ');
+}
+
+function loadFabricNames() {
```
christophertubbs, line 8: This would be a great place to get used to the `fetch` API. `$.ajax` does the job, but if you want to experiment a little, this is a great way to get used to async javascript in a small controlled manner.

python/gui/MaaS/static/common/js/domain.js
```diff
@@ -0,0 +1,206 @@
...
+        {
+            url: "fabric/names",
+            type: 'GET',
+            error: function(xhr,status,error) {
+                console.error(error);
+            },
+            success: function(result,status,xhr) {
+                if (result != null) {
+                    result['fabric_names'].forEach(function(name) {
+                        $("#fabric-selector").append("<option value='" + name + "'>" + titleCase(name) + "</option>");
```
christophertubbs, line 19: This can be refactored to look a little more like `$("#fabric-selector").append(`<option value="${name}">${titleCase(name)}</option>`);

python/gui/maas_experiment/application_values.py
```diff
@@ -173,6 +173,8 @@ def get_full_localtimezone():
 BASE_DIRECTORY = Path(__file__).resolve().parent.parent
 """The path to the base directory of the service"""
 
+STATIC_ROOT = BASE_DIRECTORY / "static"
+
 STATIC_RESOURCE_DIRECTORY = BASE_DIRECTORY / "static" / "resources"
```
christophertubbs, line 178: May as well construct this with the new definition for `STATIC_ROOT`.

python/gui/MaaS/cbv/MapView.py
```diff
@@ -142,3 +192,56 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
+            info = list()
+        else:
+            # Otherwise continue to us the passed in list
+            info = kwargs['info']  # type: list
+
+        framework_selector = datapane.Input("framework", "select", "The framework within which to run models")
+        for editor in configuration.get_editors():
+            framework_selector.add_choice(editor['name'], editor['description'], editor['friendly_name'])
+
+        pprint(framework_selector.__dict__)
```
christophertubbs, line 236: Did you mean for this debug statement to stay in here?

python/gui/MaaS/cbv/MapView.py
```diff
@@ -142,3 +192,56 @@ def get(self, request: HttpRequest, *args, **kwargs) -> HttpResponse:
...
+        """
+        The handler for 'get' requests.  This will render the 'map.html' template
+
+        :param HttpRequest request: The request asking to render this page
+        :param args: An ordered list of arguments
+        :param kwargs: A dictionary of named arguments
+        :return: A rendered page
+        """
+        # If a list of error messages wasn't passed, create one
+        if 'errors' not in kwargs:
```
christophertubbs, line 212: A list of errors and stuff wouldn't get passed in here - it would be path variables. If the path to the domain view was something along the lines of `/domain/(?P<one>...)/(?P<two>...)/(?P<three>...)`, I believe the key value pairs for `one`, `two`, and `three` would end up in kwargs. Things like list of errors, warnings, and info would come in through `request.GET`.

python/gui/MaaS/cbv/MapView.py
```diff
@@ -34,52 +36,100 @@ def _build_fabric_path(fabric, type):
...
 
 class Fabrics(APIView):
     def get(self, request: HttpRequest, fabric: str = None) -> typing.Optional[JsonResponse]:
         if fabric is None:
-            fabric = 'example'
+            fabric = 'example_fabric_name'
         type = request.GET.get('fabric_type', 'catchment')
         if not type:
-            type="catchment"
+            type = "catchment"
```
christophertubbs, line 64: It might be wise to abstract out this logic into another function that would be somewhat "easy" to swap in and out. Won't the fabrics come via the data store in the future?

python/gui/MaaS/cbv/MapView.py
```diff
@@ -34,52 +36,100 @@ def _build_fabric_path(fabric, type):
...
+
+        id_only = request.GET.get("id_only", "false")
+        if isinstance(id_only, str):
+            id_only = id_only.strip().lower() == "true"
+        else:
+            id_only = bool(id_only)
```
christophertubbs, line 70: `bool("false") == True`

christophertubbs: There is a helper function in `dmod.core.common` that can help identify truthy values from a string. I believe it's called `is_true`.

python/gui/maas_experiment/application_values.py
```diff
@@ -173,6 +173,8 @@ def get_full_localtimezone():
 BASE_DIRECTORY = Path(__file__).resolve().parent.parent
 """The path to the base directory of the service"""
 
+STATIC_ROOT = BASE_DIRECTORY / "static"
```
christophertubbs, line 176: Has this been tested inside and outside of docker? Setting Django's `STATIC` variables can end up having odd or ill or differing effects when placed behind nginx due to routing issues.

python/gui/MaaS/templates/maas/domain.html
```diff
@@ -0,0 +1,118 @@
+{% extends 'base.html' %}
```
christophertubbs, line 1: May want to inherit from `master.html` instead.

python/gui/maas_experiment/settings.py
```diff
@@ -34,6 +35,8 @@
 # packet sniffer and use the cookie to hijack the user’s session.
 SESSION_COOKIE_SECURE = not DEBUG
 
+CSRF_TRUSTED_ORIGINS = os.environ.get('TRUSTED_ORIGINS', '').split(',')
```
christophertubbs, line 38: Has this been tested behind docker with and without a cert and local vs remote? Playing with origins may cause issues when working between machines on different hosts or when working locally outside of a docker instance.

christophertubbs:
I suggest using a grid display with the height filling the entire viewable area within the containing box, with the hydrofabric fieldset in row 1 through column 3, the catchment select in row 2 column 1, the buttons for selecting what to move around in row 2 column 2, and the selected catchments in row 2 column 3. Either that or a flex box will make it easier to take advantage of your viewable area.

christophertubbs:
@robertbartel @aaraney 

Is there a name for the front end pattern of having one box of values that may move over to the other that's in this PR? It's ultra common, so I wonder if there's a simple implementation that we can import for use.

python/gui/MaaS/static/common/js/domain.js
robertbartel: This derived directly from a similar function in _map.js_.  I don't really want to perfect the implementation at this stage, especially while trying to apply something considerably different (that I'm already not especially familiar with).  I'm going to leave this alone unless you feel like there is a really compelling reason to try to do this now.

python/gui/MaaS/static/common/js/domain.js
robertbartel: I've gone through and changed the multi-variable `var`s to single variable `let`s, along with a few small tweaks to variable names.

python/gui/MaaS/static/common/js/domain.js
christophertubbs: Hey, I wrote that _before_ I knew how to use `fetch`!

>  I don't really want to perfect the implementation at this stage, especially while trying to apply something considerably different (that I'm already not especially familiar with)

Great decision. Can you add a @TODO near it or a ticket for converting traditional ajax calls to the `fetch` api? Chances of addressing the TODO are probably low, but at least it'll be in there as low hanging fruit for anyone that might be interested.

python/gui/MaaS/static/common/js/domain.js
christophertubbs: Feel comfortable adding some jsdoc? I understand if you're running to get all this in there so you can move on, but if you feel it's reasonable, jsdoc entries help with stuff like type hinting and will make it easier for others to work on this in the future.

PR #701, opened 2024-08-16
Fix bug with --node-count arg for worker.

robertbartel:
Fixes a small bug found during testing of #675, but an independent issue unto itself.

The changes allows for the _Launcher_ object to determine the correct value for the `--node-count` arg passed to job worker entrypoint script when starting services for a job.  Prior to this change, the _Launcher_ obtains this value simply by counting allocations.  With this change, it will correctly use the sum of the `cpu_count` values for all allocations for the job.


christophertubbs (approved):
You had me at "one assignment per line"

PR #728, opened 2025-07-02
Create pyproject.toml for deployment improvements

ZacharyWills:
[Short description explaining the high-level reason for the pull request]

## Additions

-pyproject.toml

Created from the requirements.txt using `uv`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,75 @@
+[project]
+name = "dmod"
+version = "0.1.0"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
+    "aiohttp~=3.8",
+    "asgiref>=3.8.1",
+    "attrs>=25.3.0",
+    "autopep8>=2.3.2",
+    "certifi>=2025.6.15",
+    "cffi>=1.17.1",
+    "channels>=4.2.2",
+    "channels-redis>=4.2.1",
+    "click>=8.2.1",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -0,0 +1,75 @@
+[project]
+name = "dmod"
+version = "0.1.0"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.12"
```
robertbartel, line 6: So this is the only thing that give me pause.  The previous project standard was to maintain compatibility with currently supported Python versions.  I think we could bend that rule and remove 3.9, given how close it is to EOL, but 3.10 and 3.11 still each have a decent amount of time left.   

We could discuss changing the project standard too, but I suspect unless something in this compels us to move to >= 3.12, it's simpler just to make a minor adjustment in this PR.

```suggestion
requires-python = ">=3.10"
```

robertbartel (commented):
Thanks @ZacharyWills, this looks really good.  I did have one thing I was worried about regarding Python version support that we probably should adjust unless moving to 3.12 is strictly required for this.

robertbartel:
I don't mind moving the whole project to 3.12.  There are just other places that would indicate earlier versions are supported, so this would be adding another minor inconsistency.  But Chris makes a good point, so as long as cleaning up that inconsistency doesn't get overlooked (i.e., it gets tracked with an issue), moving to 3.12 is probably the better long-term decision.

robertbartel (approved):

